PR #8, opened 2020-08-17
Edits to HotRegion.py and Instrument.py

DevarshiChoudhury:


xpsi/Instrument.py
```diff
@@ -213,6 +213,14 @@ def energy_edges(self):
...
 
+        try:
+            assert energy_edges.ndim == 1
+            assert (energy_edges >= 0.0).all()
+            assert energy_edges.shape[0] == self._matrix.shape[1] + 1
+            assert not (energy_edges[1:] <= energy_edges[:-1]).any()
+        except AssertionError:
+            raise EdgesError('Energy edges must be in a one-dimensional '
+                             'array, and must be postive and increasing.')
         if not isinstance(energy_edges, _np.ndarray):
```
ThomasEdwardRiley, line 224: you should check that it's an ndarray first, before trying to lookup attributes like `ndim`.

xpsi/Instrument.py
```diff
@@ -213,6 +213,14 @@ def energy_edges(self):
     def energy_edges(self, energy_edges):
         """ Set the energy edges in keV. """
 
+        try:
```
ThomasEdwardRiley, line 216: try-except blocks like this are ugly, but remnant of what I chose a while ago to reduce LOC. Is there a more elegant way than asserts in try-except block?

xpsi/HotRegion.py
```diff
@@ -265,8 +265,6 @@ def __init__(self,
         The phase of the hot region, a periodic parameter [cycles].
         """
         phase_bounds = bounds.get('phase_shift', None)
-        if not phase_bounds or None in phase_bounds:
```
ThomasEdwardRiley, line 268: have you tried executing code with this change?

I do not believe this fixes the problem because `phase_bounds` is assumed to be an iterable in the next check.

CHANGELOG.rst
```diff
@@ -17,6 +17,9 @@ Summary
 
 Fixed
 ^^^^^
+* Removed the ``ValueError`` raised by :attr:`~.HotRegion.__init__` ``if not phase_bounds or None in phase_bounds`` so that user can set a derived ``phase_shift`` value.
```
ThomasEdwardRiley, line 20: if you view this file formatted, there will be some Sphinx issues with the cross-referencing I believe.

also general tidying like adding punctuation.

ThomasEdwardRiley, line 20: if you format this file with Sphinx by rebuilding the doc pages, I believe you will find some problems with the cross-referencing to the API.

also general tidying like adding add necessary punctuation would be appreciated for readability.

xpsi/HotRegion.py
DevarshiChoudhury: Right, that also needs to be tended to. I am yet to re-run the ST-S model with changes to HotRegion.py. Having made the changes listed in this pull-request, I had re-run the ST-U model which worked but that didn't rely on this bug.

ThomasEdwardRiley:
Closed by patch v0.5.4.

PR #16, opened 2021-09-22
Updating the HotspotRegion  figure and synthetize class

yveskini:
Updating the HotspotRegion figure in the documentation ( figure made by Serena) and synthesize class to handle zero background and make sure that the input background does not get modified once it gets called more than twice ( Tuomo and I).


ThomasEdwardRiley:
Included in patch v0.7.8. Thanks!

Issue #72, opened 2022-09-19
Create script to switch between Numerical and Blackbody


sguillot:
copy files (`hot', 'elsewhere') from `surface_radiation_field/archive/' into 'surface_radiation_field' and recompile.

Basdorsman:
We should discuss this (or discuss with Yves), since a script exists in at least my fork.

sguillot:
Ah, thanks @Basdorsman 

I was thinking about including this into `setup.py` file as an option, since everything needs to be recompiled anyway. But I'm not too familiar with 'setup.py' configurations.
Thoughts ?

sguillot:
Done with #102

PR #102, opened 2022-09-19
Updating setup.py

sguillot:
Addressing issue #72:

Adding optional flags to copy the contents of /hot/numerical.pyx or /elsewhere/numerical.pyx, in order to compile XPSI with numerical atmosphere.

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -325,20 +325,17 @@ compilation in the ``setup.py`` script.
 
 .. note::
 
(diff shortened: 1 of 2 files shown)


thjsal (commented):
I checked that this works (re-installing the extensions back and forth) and it seems a good solution for me. I would just add a mention of these options in the installation documentation pages (replacing the text I lately added there about switching between different atmosphere extensions).

Basdorsman:
Great work! I propose we change the names to "NumericalHot", "NumericalElse" I have a new atmosphere type in the pipeline which would be something like "ComptonHot" but that is not finished yet. This would just be good for future proofing.

PR #103, opened 2022-09-19
Fixes #58

### LINKED ISSUE #58: Matplotlib imported in Photosphere.py even though it would not have to
The try-except-else clause checking if Matplotlib can be imported in `xpsi/Photosphere.py` is currently useless since it never gives any exception as `_mpl` is always set to `None` in `xpsi/global_imports.py` if no matplotlib can be imported and `try: None` does not raise any exception:

https://github.com/xpsi-group/xpsi/blob/6b7c8d8430fb11716f0469e7dd16ec45ba3fd7c9/xpsi/Photosphere.py#L21

Suggested solution:
Change
```
try:
    _mpl
except NameError:
    pass
else:
    import matplotlib
    from matplotlib import pyplot as plt
    from matplotlib.figure import Figure
    from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
    from matplotlib import rcParams
    from matplotlib.ticker import MultipleLocator, AutoLocator, AutoMinorLocator
    from matplotlib import gridspec
    from matplotlib import cm
    from matplotlib import animation
    import matplotlib.image as mgimg
```
to
```
if _mpl is not None:
    import matplotlib
    from matplotlib import pyplot as plt
    from matplotlib.figure import Figure
    from matplotlib.backends.backend_agg import FigureCanvasAgg as FigureCanvas
    from matplotlib import rcParams
    from matplotlib.ticker import MultipleLocator, AutoLocator, AutoMinorLocator
    from matplotlib import gridspec
    from matplotlib import cm
    from matplotlib import animation
    import matplotlib.image as mgimg
else:
    pass

```
Fixing this should allow inference runs in clusters without having the need for installing and importing Matplotlib.

thjsal:
Allowing production runs without importing matplotlib.

### DIFF
--- a/xpsi/Photosphere.py
+++ b/xpsi/Photosphere.py
@@ -18,11 +18,7 @@ from .pixelmesh.integrator import integrate as _integrate
 from .tools.energy_integrator import energy_integrator
 from .tools.phase_integrator import phase_integrator
 
-try:
-    _mpl
-except NameError:
-    pass
-else:
+if _mpl is not None:
     import matplotlib
     from matplotlib import pyplot as plt
     from matplotlib.figure import Figure
@@ -33,6 +29,8 @@ else:
     from matplotlib import cm
     from matplotlib import animation
     import matplotlib.image as mgimg
+else:
+    pass
 
 class Photosphere(ParameterSubspace):
     """ A photosphere embedded in an ambient Schwarzschild spacetime.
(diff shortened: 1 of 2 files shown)


thjsal:
I tested that this works for me (=matplotlib is not imported and nothing essential crashes) if I manually turn the matplotlib off in the global_imports.py or if I run using many cores, indicating an inference run, as the idea in the code seems to be.

sguillot (approved):
I think it looks good. I modified the warning in `global_imports.py` in case matplotlib is not found.

yveskini (approved):
It looks fine!

sguillot (approved):

PR #108, opened 2022-09-20
fixed_spin method deleted from Spacetime class

sguillot:
* removing fixed_spin 
* modification of the tutorial notebook mentioning fixed_spin

### DIFF
--- a/docs/source/model_construction_streamlined.ipynb
+++ b/docs/source/model_construction_streamlined.ipynb
@@ -137,7 +137,7 @@
    "cell_type": "markdown",
    "metadata": {},
    "source": [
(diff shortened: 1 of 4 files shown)


sguillot:
Can someone with a working sphinx compile the documentation to check that the three modified ipynb look good ?

thjsal:
I can still check the compilation. But is there a reason why you changed the mass prior from (1.0,3.0) to (None,None)? The latter puts the lower limit of the mass to 0.001 instead of 1.0. And you provided values for the space time parameters, when previously only the value for frequency was given? Would these changes affect the results shown in the tutorials for sampling e.g.?

thjsal:
`And you provided values for the space time parameters, when previously only the value for frequency was given?`

Sorry, I see now that is actually not the case, since you created the Spacetime object again with value dictionary including only the frequency. So only question is then why (None,None) instead of (1.0,3.0)?

thjsal (approved):
Looks good now!

PR #116, opened 2022-09-21
Tests for Spacetime class

dhuppenkothen:
This patch contains a new script `test_spacetime.py` in a new folder `tests` within the XPSI code folder. This contains tests for the `Spacetime` class, using the pytest package for testing. I've also added a line to the docs to mention that for running the tests, the pytest package is required. 

Once these tests are merged, I can start working on getting the automated testing infrastructure to work :)

### DIFF
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -125,6 +125,9 @@ The following Python packages are required for nested sampling:
     personal machine to gain experience on application to inexpensive test
     problems. Below we offer `from source`__ instructions.
 
+Running the tests requires:
+* `Pytest <http://pytest.org>`_
+
 The following Python packages are required for full functionality of the
 post-processing module:
 
--- a/xpsi/Spacetime.py
+++ b/xpsi/Spacetime.py
@@ -39,6 +39,10 @@ class Spacetime(ParameterSubspace):
 
     def __init__(self, bounds, values):
 
+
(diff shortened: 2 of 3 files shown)


sguillot:
The test method `test_fails_when_input_is_not_dict` crashed with error:
```
        with pytest.raises(TypeError):
>           st = Spacetime(bounds, self.values)
 ...
 ...
E       AttributeError: 'tuple' object has no attribute 'get'
```
I fixed the test method by replacing `with pytest.raises(TypeError):` by `with pytest.raises(AttributeError):`

Does that make sense ?

sguillot:
The names of two test methods are switched.
For the method below, you give `Spacetime(self.values)` so the method name should be `with _bounds_missing`
```
  def test_spacetime_breaks_with_values_missing(self):
        with pytest.raises(TypeError):
            st = Spacetime(self.values)
```

For the method below, you give `Spacetime(self.bounds)` so the method name should be `with _values_missing`
```
    def test_spacetime_breaks_with_bounds_missing(self):
        with pytest.raises(TypeError):
            st = Spacetime(self.bounds)
```

sguillot:
This looks all good to me and ready to be merged. 
Are we trying to get the a and q calculation tests working this week, or keep that for later ?

sguillot (approved):

PR #117, opened 2022-09-21
Clarifying errors in instrument class and variable 'array' name change

sguillot:


thjsal (approved):
Approved after the changes you were just planning to make.

PR #142, opened 2022-09-22
FAQs and common problems document updated from team Overleaf

drannawatts:


thjsal (approved):
I cleaned the common problems quite a bit and left some of the problems out (those that appeared a bit less familiar to me or had no time to clean). Fine for me to merge.

PR #145, opened 2022-09-22
Added utilities directories with ImportanceSample.py and ProjectionTool.py

serenavinciguerra89:


thjsal:
Can you add  `return ax` at the end of the Projection tool code? Then the users can further modify the figure if they want.

serenavinciguerra89:
Do you see the added return ax now? or do I have to Tell Git that I want a pull request with that commit in as well?

sguillot:
We see it :-)

thjsal:
At some point, you also created an option to have an antipodal projection of one of the spots. Do you think that could be added to this version as well?

serenavinciguerra89:
I had completely forgotten about that, sorry…
I need to check how I did that and if it is easy to transfer it, I’ll do it
(I need to check if I modified the file directly without having put it as
an option or if I added the option)
I’ll check that tomorrow.

thjsal:
No worries. Now I also realize this version we still have "Nord pole" instead of "North pole" :). I suggest to fix that as well.

serenavinciguerra89:
I cannot see the indentation from the github (I thought I solved those on
Wednesday and now in my version I checked and the indentations of those
lines seem fine). Added the __init__.py and the line in the setup.py

Il giorno ven 14 ott 2022 alle ore 23:35 Tuomo Salmi <
***@***.***> ha scritto:

xpsi/utilities/ProjectionTool.py
```diff
@@ -0,0 +1,648 @@
...
+        for i in range(len(LABall)):
+            if (OMIlab in LABall[i]):
+                coli = 'black'
+            else:
+                ind_i = ind[argT[j]]
+                coli = mycolors0[ind_i]
+                j = j+1
+            CAall.append(coli)
+            
+    if antipodal:
```
thjsal, line 291: I think here is at least an indentation error. There should be 4 more spaces before `if antipodal`.

serenavinciguerra89:
They do not seem to have a weird indent on my version (also in the code
lines you printed it is not clear if it has a problem with indent as I
think it should align with the else above those lines... see attachment).
I also tested it in one post-processing and I could import the file without
problems and use it normally. I also checked the antipodal option, and it
works as far as I can tell.


Il giorno sab 15 ott 2022 alle ore 19:06 Tuomo Salmi <
***@***.***> ha scritto:

thjsal:
I don't see any attachments But in case that `if antipodal` is aligned correctly, the problem is that the lines following it are double intended (they are 8 spaces more intended than `if antipodal` at the moment, right? Or is your version the same as currently in GitHub?

thjsal:
To summarize: The indentation problem above was solved, bu the CI test still fails, since it apparently tries to read the Projection Tool (for some reason) as if that would need to be written in Python3 (although that is not yet required).  
Regardless of that, I think we could proceed to merge this. However, I think we should update the CHANGELOG-file with this, and update X-PSI version numbers to e.g. 1.0.1.

thjsal (approved):

PR #146, opened 2022-09-22
Ensuring that data and background spectrums have the same shape

yveskini:


sguillot (commented):
Just one quick comment: np_ is already imported from global-imports as _np.

I have tested the code yet.

thjsal (approved):
Approving this change (just removed the unnecessary imports)  although based on my testing  this new check seems to still require that one sets `support=support `when calling the original init-function in the init-function of CustomSignal. Otherwise, limits were not checked.

PR #148, opened 2022-09-22
Limitation on number of plots set to 9 due to prevent plot overcrowding.

### LINKED ISSUE #56: Checking possibility of allowing more than 5 `IDs` to be post-processed at the same time.
https://github.com/xpsi-group/xpsi/blob/29ead0d62c4c0697ad35688d9d0b29b083f11e12/xpsi/PostProcessing/_postprocessor.py#L153

The above code and a few other following lines in the same function seem to deliberately limit the number of `IDs` to be post-processed to be less than or equal to 5. The objective is to allow as many as the user desires if possible.

yveskini:
Fixes #56

### DIFF
--- a/xpsi/PostProcessing/_postprocessor.py
+++ b/xpsi/PostProcessing/_postprocessor.py
@@ -150,12 +150,12 @@ class PostProcessor(object):
                                 'collections.OrderedDict container.')
 
         if len(self._subset) > 1:
-            if len(self._subset) > 5:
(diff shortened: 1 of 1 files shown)


thjsal (commented):
Should you have also changed to word "five" to "nine"?

DevarshiChoudhury (commented):
All looks fine. Tuomo's comment's also been accounted for.

PR #153, opened 2022-09-23
Added badges to readme

dhuppenkothen:
Test PR to see if it triggers CI. Also adds some sweet badges to the README


DevarshiChoudhury (commented):
Hyperlinks need to be fixed. Incorrect link for the ``pull`` page, points to stingray instead of X-PSI

PR #174, opened 2022-10-18
Build docs via GitHub Actions

dhuppenkothen:
... so that you don't need to do this manually. Not sure it's working yet.


dhuppenkothen:
ping @drannawatts @thjsal. I think this is ready for merging if someone wants to do a code review. The last change will reverse the changes to the documentation I made, and will also set it up to run this workflow on merging a branch into main.

thjsal (approved):
Looks good to me.

PR #177, opened 2022-11-09
Changelog updated and version changed to v1.1.

thjsal:


thjsal:
Before actually merging this, I can add a comment in the Changelog about the new automatic documentation compilation using GitHub actions that is also now there (hopefully working).

PR #179, opened 2022-11-11
Moving X-PSI to Python3

dhuppenkothen:
Work-in-progress pull request to move X-PSI to Python3.


drannawatts (approved):
"It will probably be fine" a wise woman said.

PR #183, opened 2022-12-02
New numerical atmosphere extension for free beaming runs

thjsal:
A new numerical atmosphere extension, allowing a parametrized atmospheric beaming correction, has been added to the surface radiation field archive, and an option in the setup.py file has been added to be able to install X-PSI using that. 

Example scripts and/or instructions in the tutorials will still be added to demonstrate how to use this extension.


thjsal:
An example script has now been added, and it is also mentioned and linked in the documentation.

drannawatts (approved):
Changes look good to me.

PR #186, opened 2022-12-09
Updated licence info and development status in setup.py file

drannawatts:
Updated lience information and development status in the setup.py file

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -273,7 +273,7 @@ if __name__ == '__main__':
         classifiers = ['Development Status :: 5 - Production/Stable',
                        'Intended Audience :: Science/Research',
                        'Operating System :: Linux, macOS',
-                       'License :: GPLv3',
+                       'License :: OSI Approved :: GPLv3',
                        'Programming Language :: Python'],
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -270,10 +270,10 @@ def EXTENSION(modname):
         include_package_data = True,
         ext_modules = extensions,
         cmdclass = cmdclass,
-        classifiers = ['Development Status :: 3 - Alpha',
+        classifiers = ['Development Status :: 5 - Production/Stable',
                        'Intended Audience :: Science/Research',
                        'Operating System :: Linux, macOS',
-                       'License :: OSI Approved :: MIT License',
+                       'License :: GPLv3',
```
thjsal, line 276: Seems to me that GPLv3 is also an OSI approved license :https://opensource.org/licenses/alphabetical. 
So we could still have "OSI Approved" mentioned there?

drannawatts: Agreed, have just put it back in.

thjsal (approved):
Looks good to me!

PR #191, opened 2022-12-12
Allowing user to decide if/when likelihood calculation should be fully computed.

thjsal:


DevarshiChoudhury (approved):
This is a good idea! 😀 
Implementation seems fine to me 👍

DevarshiChoudhury:
The header says [WIP]. Should we expect more changes or shall I merge it?

yveskini (approved):

Issue #170, opened 2022-10-12
Use standard GetDist and NestCheck (or convert customised versions for Python3)


thjsal:
Currently we use the customzied versions of GetDist and NestCheck packages in postprocessing:
https://github.com/ThomasEdwardRiley/getdist/tree/customisation
https://github.com/ThomasEdwardRiley/nestcheck/tree/feature/getdist_kde

For Python3 version, these should be updated (need a new fork?), or we should modify X-PSI to survive without them being customized.

thjsal:
For example, if using a matplotlib version that is not too old for Python3, I get this error:

```
plots.py:10, in <module>
      7 import six
      8 import warnings
---> 10 matplotlib.use('Agg', warn=False)
     11 from matplotlib import cm, rcParams
     12 import matplotlib.pyplot as plt

TypeError: use() got an unexpected keyword argument 'warn'
```

dhuppenkothen:
What is actually different in those versions from the official packages? Is there some way we could make xpsi use the actual packages? That would probably be less fragile in the long run.

thjsal:
I am not very familiar with these customizations but according to Riley+2019 paper (footnotes 90 and 91) this GetDist version has some minor plotting customization, and the NestCheck version is customized to support the use of GetDist KDE. 

I don't remember exactly which things break if using the official packages, but at least some functionalities in X-PSI post-processing used to result in errors if not using these versions. That could be checked again still.

thjsal:
So the plan is now to update X-PSI so that it could use the standard NestCheck and GetDist packages when moving to Python3. 

The first problem that encounters is related to file namings, giving e.g.
`FileNotFoundError: ./Outputs/ST_live_1000_eff_0.3_seed0_transformed-dead-birth.txt not found.
` ,
when using the standard NestCheck (and running the PostProcessing example in https://github.com/xpsi-group/xpsi/blob/main/examples/examples_fast/Postprocessing.ipynb). The file name that X-PSI produces ends with `transformeddead-birth.txt`.  I did not test it yet, but might be solved by changing how filenames are appended in https://github.com/xpsi-group/xpsi/blob/main/xpsi/PostProcessing/_backends.py for transformed parameters.

thjsal:
Okay, seems like the solution for this problem is just to replace                        
` _np.savetxt(filerootpath + ext, transformed)`
 with
`_np.savetxt(filerootpath + "-" + ext, transformed)`
[...]

PR #192, opened 2022-12-14
Updating X-PSI py3 for standard getdist and nestcheck.

thjsal:


thjsal:
I pushed the fixes found to the problems solved so far when converting X-PSI to be compatible with standard GetDist and NestCheck version, which were discussed in this issue: #170 

More problems still need to be solved though.

thjsal:
Issues mentioned before got solved, but new was found:
https://github.com/xpsi-group/xpsi/issues/170#issuecomment-1384416831

thjsal:
I removed some previously commented or not-used-anymore lines from the `_corner.py` (which need to be put back if later still want to change to the customized nestcheck functionalities)  and expanded the PostProcessing script with pulse and spectral plots. I also copied the new notebook the the documentation pages (since still did not manage to compile documentation properly when reading it from a different path). 

Please, approve, when you think it is ready. Or suggest what else should be changed.

yveskini:
Everything is working fine so far.

yveskini (approved):

Issue #187, opened 2022-12-09
Add Multinest LD_LIBRARY_PATH instructions to main install page


drannawatts:
In order to run correctly, the user needs to export the following path (or something similar depending on where MultiNest was installed):
export LD_LIBRARY_PATH=$LD_LIBRARY_PATH:$HOME/multinest/MultiNest_v3.12_CMake/multinest/lib/

At present this instruction is only on the HPC installation page, but it needs to be added to the main installation page as well.

PR #193, opened 2023-01-10
Minor corrections to installation documentation

drannawatts:
Added the line about setting the environment variable for the library path to point to Multinest, which was previously on the HPC instructions page, to the main installation page.

Corrected some minor typos. 

Addresses issue #187

### DIFF
--- a/docs/source/HPC_systems.rst
+++ b/docs/source/HPC_systems.rst
@@ -141,7 +141,7 @@ Use the last command to check for the presence of shared objects.
 
 .. note::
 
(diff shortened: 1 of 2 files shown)


serenavinciguerra89:
@drannawatts did you also add the lines of the library to the Helios installation? As I could not see that highlighted

thjsal:
Yes, unless other people think those lines would be useful/important in the Helios instructions too, I think it is actually fine like this.

thjsal (approved):

PR #198, opened 2023-01-27
Added option to fix the random seed for the synthetic data generation with python3 version.

thjsal:


thjsal:
I also modified the example notebook generating synthetic data, and found that fixing the seed to 42 gives exactly the same data (based on the look of the residual plot) as it gave for python2 version.

dhuppenkothen (approved):
This looks good to me. Quick check: is `synthesise.pyx` used by code somewhere else in X-PSI, where we should add explicit references to the new `gsl_seed` parameter?

thjsal:
A reference to this was added to the Synthetic_data.ipynb notebook, where this is used. This not is still not mentioned in the Modeling tutorial of the internet pages yet (where synthetic data is also produced), but I think we can later add it there as well.

PR #223, opened 2023-01-31
Noopenmp install option

sguillot:
I added a `noopenmp` option in the `setup.py` that will remove the `fopenmp` compilation arguments.
@drannawatts, could you give it a try ?


drannawatts (approved):
Works for me just fine.  Need to add this to the installation docs description for Mac.

PR #233, opened 2023-02-01
Moving conda install commands to one line to speed up build

dhuppenkothen:
Following Johannes Buchner's suggestion, trying to speed up GH Actions.


dhuppenkothen:
This is ready for review. It changes the Github Actions to use mamba (which is kind of a faster version of conda) and moves install commands to the same line. It also introduces caching of the environment, which should hopefully also speed up the tests + building documentation

thjsal (approved):

PR #241, opened 2023-02-01
HPC installation instructions for Snellius+Helios updated for python3

thjsal:


thjsal:
Example jobs are now added, and this pull request should be ready for a full review.

Basdorsman (approved):
I tested the helios installation and the helios example job and they work!

PR #244, opened 2023-02-01
Six changes for Py2 to Py3

drannawatts:
Changes made in response to suggestions from Six command 2to3 to modify the code from Python 2 to Python 3.  

Please check carefully!  Some seem obviously sensible, others are a bit more obscure to me.


xpsi/PostProcessing/_cache.py
```diff
@@ -91,7 +91,7 @@ def __next__(self):
 
         with self._open('r') as f:
             g = f['data']
-            for key in g.keys():
+            for key in list(g.keys()):
```
dhuppenkothen, line 94: I don't think you need to turn this into a list here, do you? `g.keys()` should be a generator that you can step through in a for-loop? Might be worth checking.

xpsi/PostProcessing/_corner.py
```diff
@@ -497,7 +497,7 @@ def _plot_triangle(self,
         if scale_attrs is None:
             scale_attrs = {}
 
-        for key, value in scale_attrs.items():
+        for key, value in list(scale_attrs.items()):
```
dhuppenkothen, line 500: Same here, IIRC part of why these methods were turned into generators is so that you don't *have* to allocate space in memory for the entire list object. So turning them into lists might defeat the purpose? I might be wrong, though.

examples/examples_fast/Modules/main.py
```diff
@@ -8,10 +8,10 @@
 
 import time
 
-from CustomInstrument import CustomInstrument
-from CustomSignal import CustomSignal
-from CustomPhotosphere import CustomPhotosphere
-from CustomPrior import CustomPrior
+from .CustomInstrument import CustomInstrument
```
sguillot, line 11: It doesn't work with the `.` in front of the Custom class imports. I get two types of errors. Either:
```
  File "main.py", line 11, in <module>
    from .CustomInstrument import CustomInstrument
ValueError: Attempted relative import in non-package
```
or 
```
  File "/Users/sebastien/Research/XPSI_RUNS/PY3_TESTS/examples_fast/Modules/main.py", line 11, in <module>
    from .CustomInstrument import CustomInstrument
ImportError: attempted relative import with no known parent package
```
Depending on whether I try to run the main from the original xpsi directory, or copy the examples directory elsewhere and try to run the main.

It works fine when I remove the `.`

dhuppenkothen (changes requested):
looks good overall. it's worth checking on the behaviour of generators (e.g. zip()) when iterating through a for-loop compared to when assigning it to a variable, I think, before merging.

xpsi/EnsembleSampler.py
```diff
@@ -172,7 +172,7 @@ def _nd_ball(self, moments):
         else:
             raise PriorError
 
-        moments = list(map(list, zip(*moments)))
+        moments = list(map(list, list(zip(*moments))))
```
sguillot, line 175: Both `list(map(list, zip(*moments)))` and `list(map(list, list(zip(*moments))))` work.
@dhuppenkothen Do you have a sense of which should be use ?

xpsi/PostProcessing/_metadata.py
```diff
@@ -43,7 +43,7 @@ def __init__(self, ID, names, bounds=None, labels=None,
             self.bounds = bounds
 
         if labels is None:
-            self._labels = dict(zip(self.names, self.names))
+            self._labels = dict(list(zip(self.names, self.names)))
```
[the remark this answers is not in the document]
sguillot: Actually, in python3, it seems that `dict(list(zip(....)))` returns an empty dictionary, but `dict(zip(....))` returns a proper dictionary.  In python2, both return the same thing.

examples/examples_fast/Modules/main.py
dhuppenkothen: I am a strong supporter of absolute rather than relative imports in packages in general, though I'm not sure how that would work here.

xpsi/PostProcessing/_metadata.py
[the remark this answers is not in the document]
sguillot: Python3 on the left, Python2 on the right
dict list zip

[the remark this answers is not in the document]
sguillot: So, I suggest reverting to ` self._labels = dict(zip(self.names, self.names))`

xpsi/utilities/ImportanceSample.py
```diff
@@ -26,31 +26,31 @@
 args = parser.parse_args()
 
 if not args.path_to_mains:
-    args.path_to_mains = input('Specify Absolute or relative path to main files: ')
+    args.path_to_mains = eval(input('Specify Absolute or relative path to main files: '))
```
sguillot, line 29: `eval` should not be used for string inputs (paths, filenames, etc.) because it will try to search for a variable with the name of the input string.

xpsi/utilities/ImportanceSample.py
```diff
@@ -26,31 +26,31 @@
...
-    names = input('Specify the name of the model variables: ') #check if input
+    names = eval(input('Specify the name of the model variables: ')) #check if input
     args.names = names.split()
 else:
     args.names = args.names[0].split()
 
 def check_prior_change(test):
     if not args.prior_change:
-        args.prior_change = input('Specify if importance sampling changes priors of model variables  (True/False): ')
+        args.prior_change = eval(input('Specify if importance sampling changes priors of model variables  (True/False): '))
```
sguillot, line 53: the `eval` can be left here (assuming the user does write "True" or "False"), but it also works without.

sguillot (changes requested):
See comments for requested changes.
The rest works well.

drannawatts:
OK so with corrected commits essentially the combined effect was to remove a lot of 'from future' statements and correct some unicode stuff :-D

sguillot (approved):

dhuppenkothen (approved):
👍 LGTM

PR #246, opened 2023-02-01
Py3 redo the intro pages install page 205

Basdorsman:
Installation instruction page for Python 3 is significantly streamlined. Updates to Multinest installation. Updates to Mac OS tips. Basic_environment.yml is now environment.yml.


Basdorsman:
I have reintroduced "basic_environment.yml" because without it the CI tests fail. Tuomo also mentioned he is using it for installation on HPC systems. I figured we can keep it for now but it would be more streamlined if we have fewer environment.yml files.

Basdorsman:
I'm checking now if the full installation works, also given that I am using a JohannesBuchner Multinest version rather than FarhanFeroz version as in the old installation instructions.

sguillot (changes requested):
- There is a broken RST link to Multinest in `which may be installed later (refer to ...`
- Change the reference to `gsl-latest/build` to something like "Enter the untared directory and type ./configure...

Basdorsman:
add path for gsl also

yveskini:
Removing the Post-processing packages in the environment makes its creation much faster. 
 The removed  Post-processing packages could simply be install by conda-forge.

Basdorsman:
@sguillot would you be available to do a final review before merge?

thjsal:
In case worth to mention here, I installed emcee using this command: `conda install -c conda-forge emcee`

sguillot:
I will implement some small changes.

sguillot:
@Basdorsman  if you find instructions to install BLAS, LAPACK, etc, can you add them ?

The rest is good for merge I think

sguillot (approved):

PR #255, opened 2023-02-03
General py2 documents update

drannawatts:
Updates to Citation, Team & Acknowledgements, Applications pages.

Update to Future page (for py2, to point out that it has none).

Note:  will break some cross-references to papers (R19, R21 references) in the py2 documentation.  Not fatal but may fix this at a later date.

### DIFF
--- a/docs/source/todo.rst
+++ b/docs/source/todo.rst
@@ -4,5 +4,5 @@ Future
 ------
 
 All good things come to an end and sadly there is no future for the Python 2 variant
-of X-PSI.  Please return to the main X-PSI page and switch to the Python 3 (X-PSI v>2.0)
-variant and see what the future holds! 
+of X-PSI.  Please return to the `main X-PSI page <https://xpsi-group.github.io/xpsi/>`_
+ and switch to the Python 3 (X-PSI v>2.0) variant to see what the future holds! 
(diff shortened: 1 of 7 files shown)


docs/source/todo.rst
```diff
@@ -3,16 +3,6 @@
...
-* Moving to Python 3 from Python 2 (next major X-PSI release)
-* Providing examples for importance sampling
-* Documenting new post-processing features and improved visualisation tools.
-* Additional features to support parameter recovery tests
-* New atmosphere extension modules
-* Support for X-ray polarimetry 
-* Automatic module generator
-
+All good things come to an end and sadly there is no future for the Python 2 variant
+of X-PSI.  Please return to the main X-PSI page and switch to the Python 3 (X-PSI v>2.0)
```
dhuppenkothen, line 7: Should we turn `main X-PSI page` into a link to the default docs of X-PSI (which should now be Python3)?

dhuppenkothen (approved):
Small suggestion, but I'll approve it pre-emptively :)

docs/source/todo.rst
drannawatts: Oh yes good idea. What link would that be?

dhuppenkothen:
This one: https://xpsi-group.github.io/xpsi/

docs/source/todo.rst
dhuppenkothen: This one: https://xpsi-group.github.io/xpsi/

PR #257, opened 2023-02-03
Merging python2 docs

dhuppenkothen:
I think this should merge Anna's changes in the docs into the python3 branch, if I did everything correctly.


drannawatts (approved):
So I think the changes to install.rst will be overwritten by what comes from Team New Install very shortly.

PR #269, opened 2023-02-06
[WIP] Solving the signedness problem with integers.

thjsal:


thjsal:
As mentioned in the linked issue, it seems that there are several ways to fix the integer warnings we get when installing X-PSI. In the current solution, I have several of them, depending on which seemed to be easiest in each situation. But if you have an opinion that long integers should be preferred against shorter ones, replacing `size_t` with `unsigned int` in many places was maybe not the right way to go.

thjsal:
To do: We think it is better to actually change `unsigned int`s to l`ong signed int`s, which also solves the problem.

thjsal:
Now everything should be switched to long integers instead. Instead of using `long signed int,` I used `size_t `everywhere, which correspond to `long unsigned int` (could get the former working for some reason, and the latter seemed to be anyway the type that it is mostly preferred in the code). 

Again, no warnings when installing X-PSI, and the likelihood checks give the same results as before. I think this could be ready to be merged.

DevarshiChoudhury (approved):
code changes seem fine to me and the integer warnings do not show up for me. Check [file in slack thread](https://xpsi-men.slack.com/archives/C03RF6EQLSW/p1676490092058769) for warnings produced during installation (mostly has to do with Cython fallthroughs, Mac issues and some ignored packages)

PR #280, opened 2023-02-11
Updates to JOSS paper on main branch

drannawatts:
First updating the JOSS paper on the main branch, can you double check that I have addressed all of the final comments from the Editor? 

Once this is on the main branch and I've checked that it looks OK using JOSS Editorialbot I'll do a separate pull request of the same files for the py2 branch for the tagged release.


dhuppenkothen (approved):
👍 LGTM

PR #281, opened 2023-02-11
JOSS paper updates to py2

drannawatts:
So now the same JOSS files updates, but this time to the py2 branch....


dhuppenkothen (approved):
👍

PR #284, opened 2023-02-13
Small edits for JOSS submission

dfm:
As part of the final processing, here are some small edits to the manuscript.

ref: https://github.com/openjournals/joss-reviews/issues/4977


drannawatts (approved):
Thanks for this, I had completely missed that the journal names were not there!  Which is a little horrifying.

dfm:
@drannawatts — No worries! This happens for every astro submission... we really need to just get those journal macros into out TeX template :D

PR #289, opened 2023-02-15
Python3 installation update

thjsal:


thjsal:
Installation instruction for Python3 version should be now fixed (MultiNest to be installed from correct source, and installing PyMultiNest using source instead of conda). Also fixed a couple of the small issues regarding links and missing information in some of the examples/tutorials.

sguillot:
Maybe I can add my CALMIP install instructions before we merge the PR ?
Can this wait tomorrow or Friday ?

thjsal:
Sounds fine to me! Those instructions are on a separate page, so it does not actually matter where (in which branch) and in which order we update them. But it can be combined with this if it feels easiest.

sguillot (approved):

sguillot (approved):

thjsal:
There were still plenty of warnings when compiling the documentation, but now most of them have been fixed. Especially the references to Riley+2019 in Python docstrings. They seem to work now. If all CI tests pass, I would go forward and merge this branch.

PR #294, opened 2023-03-06
Likelihood class integration test

thjsal:
A likelihood check is added as a part of the continuous integration tests. This version does not work yet, since the GItHub Actions do not seem to be able to find the correct data file (from `examples/examples_fast/Data/xpsi_good_realisation.dat`). I can bypass that by copy pasting all the data elements into the code (meaning ~ 1000 lines of extra code), which indeed worked in the second latest commit of this branch (the CI-tests passed when running them with GitHub Actions). But if someone knows a better way to do this, I would be happy to know.


thjsal:
Okay, looks like it works now! Just needed to drop the 'file:' prefix (which ChatGPT had told me to use) and know that the GitHub Actions working directory is by default the root directory of the repository. I will just write something to the ChangeLog and then it seems ready to me.

thjsal:
Okay, I added more code to be directly read from the examples instead of copying them. And made an update to the ChangeLog. But seems ok for me merging, unless someone wants to change something.

Basdorsman (approved):
works on my machine as well!

thjsal:
Thanks for checking!

PR #303, opened 2023-04-14
Atmosphere switching without re-installing X-PSI

thjsal:
Atmosphere wrapper introduced for switching between extensions without re-installing X-PSI. The implementation is not finished for all the different integrators and atmosphere+elsewhere options, but it is ready for comments. And I tested that now I can already run both the `TestRun_BB.py` and `TestRun_NumBeam.py` examples without re-installing X-PSI.


thjsal:
So the idea is to add the imports of all the atmosphere extensions (the current and any future updates) to the file called `hot_wrapper.pyx` (and perhaps separate `elsewhere_wrapper.pyx` for elsewhere), which takes an input argument to determine which atmosphere module when calculating. I made this new file to avoid repeating the atmosphere options in all the various integrators that are currently separately calling the atmosphere interpolations.  Now in all those integrators, one has to just import `hot_wrapper` instead of `hot`, and pass an argument to the `init_hot()` function specifying which atmosphere module to use.

Originally, I tried also to import only the user-specified module, but seemed not possible to do with C functions, which would all anyway get imported when compiling X-PSI, and X-PSI would just use the one that was imported last (since cannot even change the name of the functions after/while importing them). Therefore, needed to go with hard-coding the atmosphere options in X-PSI.

Basdorsman:
I would suggest we try to change the integer flags to string flags? BB, NSX, BEA, ACC?
I don't think you have implemented the passthrough of the argument from the integrator yet right?

thjsal:
Sounds like a good idea! Need to think about the names though. I did implement the passthrough for one of the integrators (the azimuthal invariant one, which is used by the example scripts), but not yet for the rest.

thjsal:
@Basdorsman Do you think string flags are enough for the user, but we could still use integer flags at the C-code level? I just updated the code to take the arguments as strings but to convert them to integers in HotRegion.py. Having integer flags at C-level seems easier for me, and perhaps also faster. However, I agree it would be then good to comment on the core level as well that which number corresponds to which model.

sguillot:
Just a note the the tutorial notebook and installation instructions will need to be updated to account for this "major" change in the functionality of X-PSI.

sguillot:
I have two suggestions:
- this new feature and the GSL fix feel significant enough to increase to version 2.1 for example ?
- should we include also the bug fix of #305 ?

thjsal:
With the GSL-fix, I actually moved X-PSI from 2.0.0 to 2.0.1 (I tagged it with 2.0.1 but did not "release"), although I forgot the mention that in the final commit message. For the next upgrade, I agree we could do both this atmosphere update, and the remaining bug fixes in the likelihood calculation, and release version 2.1.0.

xpsi/Everywhere.py
```diff
@@ -155,6 +181,32 @@ def __init__(self,
...
+                            'atmosphere options are at the moment "BB", "Num4D", and "user".')
+
+    @property
+    def beam_opt(self):
+        """ ... """
+        return self._beam_opt
+
+    @beam_opt.setter
+    def beam_opt(self,option):
+        self._beam_opt = option
```
sguillot, line 208: There should also be a`TypeError` if `beam_opt` is different from the option 0, 1, 2, 3, no ?

xpsi/HotRegion.py
```diff
@@ -690,6 +716,32 @@ def is_antiphased(self):
...
+
+    @property
+    def beam_opt(self):
+        """ ... """
+        return self._beam_opt
+
+    @beam_opt.setter
+    def beam_opt(self,option):
+        self._beam_opt = option
+
```
sguillot, line 744: There should also be a `TypeError` if `beam_opt` is different from the option 0, 1, 2, 3, no ?

xpsi/surface_radiation_field/__init__.pyx
```diff
@@ -180,33 +208,47 @@ def intensity(double[::1] energies,
...
         norm_ptr = eval_hot_norm
-    elif extension == 'elsewhere':
+    elif region_extension == 'elsewhere':
         init_ptr = init_elsewhere
         free_ptr = free_elsewhere
         eval_ptr = eval_elsewhere
         norm_ptr = eval_elsewhere_norm
     else:
-        raise ValueError("Extension module must be 'hot' or 'elsewhere'.")
+        raise ValueError("Region extension module must be 'hot' or 'elsewhere'.")
```
sguillot, line 230: Naive question:  shouldn't the `region_extension` also allow `everywhere` as well ?

xpsi/surface_radiation_field/elsewhere_wrapper.pyx
```diff
@@ -0,0 +1,55 @@
...
+                     const double *const VEC,
+                     void *const data,
+                     size_t beam_opt) nogil:
+    #Note: Beaming option for elsewhere is not implented, at least yet,
+    #even though it is one of the input parameters.
+    if atmos_extension_elsewhere == 1:
+        I_hot = eval_hot_BB(THREAD,E,mu,VEC,data)
+    else:
+        I_hot = eval_hot_Num4D(THREAD,E,mu,VEC,data)
+    return I_hot
```
sguillot, line 49: there should be `elif atmos_extension_elsewhere == 2` for Num4D, and also the option for hot_user, no ?

sguillot (commented):
I tested and it worked.
But I also wrote some comments for some of the modifications

sguillot:
Another comment:
I think we should give a notebook example of a CustomPhotosphere with the numerical atmosphere, similar to that in `TestRun_Num.py`.  Previously, users could look at the Zenodo files and scripts to duplicate, but without it, I think the "preload" function of the numerical atmosphere should be in a notebook (the `TestRun_Num` script perhaps isn't detailed enough).  Do you agree ?

CHANGELOG.rst
```diff
@@ -34,6 +34,48 @@ and this project adheres to
...
+
+Changed
+^^^^^^^
+
+* Documentation, tutorials, and example scripts updated for using the new atmosphere switching (except ``xpsi/module_generator.py``).
+
+Deprecated
+^^^^^^^^^^
+
+* The previous way of installing X-PSI with different atmosphere extensions has been deprecated. **To use the old scripts with numerical atmosphere setup, one needs to add 'atm_ext="Num4D"' as an input parameter for all the relevant HotRegion/Elsewhere/Everywhere objects.**
```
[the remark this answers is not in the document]
thjsal: I mean those Python scripts (more precisely `main.py`) that can/will be found e.g. from the Zenodo repositories of recent/upcoming X-PSI works. Perhaps should indeed clarify "old scripts" here.

xpsi/Everywhere.py
thjsal: Good point! I can add that.

xpsi/HotRegion.py
thjsal: Yes. I'll add that.

xpsi/surface_radiation_field/__init__.pyx
thjsal: I think not. There never used to be a separate atmosphere extension for everywhere (it has just used the one of hot region). The purpose of this function is to test atmosphere extensions, so testing it using "hot" already tests it for "everywhere". Maybe, if want to be super clear, we could allow region_extension=everywhere here but it would then just need to do the same as region_extension=hot.

xpsi/surface_radiation_field/elsewhere_wrapper.pyx
thjsal: I think you are right. Good catch!

xpsi/surface_radiation_field/hot_wrapper.pyx
```diff
@@ -0,0 +1,122 @@
...
+#----------------------------------------------------------------------->>>
+cdef void* init_hot(size_t numThreads, const _preloaded *const preloaded, size_t atm_ext) nogil:
+    global atmos_extension
+    atmos_extension=atm_ext
+    if atmos_extension == 1:
+        return init_hot_BB(numThreads, preloaded)
+    elif atmos_extension == 2:
+        return init_hot_Num4D(numThreads, preloaded)
+    else:
+        return init_hot_user(numThreads, preloaded)
```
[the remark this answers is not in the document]
thjsal: Yeah, probably better to do as you said. Thanks for the suggestion.

thjsal:
@sguillot Thanks for checking! I will do fixes based on your comments. Having a notebook version of `TestRun_Num.py` sounds a reasonable idea to me. But I am not sure if I understood what details you would like to add into that.

xpsi/Everywhere.py
sguillot: fixed with 7cfa468864ddf67b0c02d096ce774076b26c7607

xpsi/HotRegion.py
sguillot: fixed with 7cfa468864ddf67b0c02d096ce774076b26c7607

xpsi/surface_radiation_field/elsewhere_wrapper.pyx
sguillot: fixed with d1efbeadbeeec15b3358f9a1323ccb559aa1cc01

xpsi/surface_radiation_field/hot_wrapper.pyx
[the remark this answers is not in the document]
sguillot: Looking at it a bit more carefully, I think that we would need to decide what to do if the values are not 1, 2 or 3. Either:
- raise an error (but don't know how to do this in cython)
- or default back to BB (i.e., `else: return BB()`), but we would need to send a warning to the user...
Thoughts ?

sguillot:
> @sguillot Thanks for checking! I will do fixes based on your comments. Having a notebook version of `TestRun_Num.py` sounds a reasonable idea to me. But I am not sure if I understood what details you would like to add into that.
And even if the preload() function is there, it is not readily usable for a CustomPhotosphere

The tutorial notebook are more meant for that. But at the moment:
- the modelling notebook tutorial are for BB emission (nothing specify of the emission model in the `CustomPhotosphere`, so default to BB)
- the Surface radiation field tools notebook has the preload function, but it is just used to call the `intensity` function, and not a full CustomPhotosphere.

What would be useful is an additional notebook (or an added part in an existing notebook) showing how to define a `CustomPhotosphere` class with a generic `hot_atmosphere` loading method. When I say generic, I mean something with `size` passed as an argument (like in the `preload` of the surface radiation field tool notebook), rather than the hard-coded `size=(..,..,..,..)` that is in the `TestRun_Num.py`.

I see the `TestRun_Num.py`  as a script to test everything (without really expecting the user to dig into it). The atmosphere part of X-PSI is an important one, so I think it's crucial to provide an explicit tutorial for users to learn how to switch between BB and Num4D.  

Does that make sense ?  I'm going to work on a quick tutorial notebook showing a pulse profile for BB and one for Num4D (for all other parameters being the same). I'll commit it here on this branch

CHANGELOG.rst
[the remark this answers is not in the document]
thjsal: Clarified in a723288d29291b1ce596aedf699ed250ec6a8c37.

Basdorsman (approved):
I checked and found that the modeling tutorial examples and surface_radiation_field_tools.ipynb are working.

sguillot:
Hi @thjsal 
I made a small commits to default to BB instead of defaulting to USER in the `hot_wrapper` and `elsewhere_wrapper`.
Do you agree with this ?

I am not 100% sure about it personally, for the following reasons:

- no values other than 1, 2 or 3 should arrive there, since the checks are performed before...
- if they do (because of some oversight), it's unclear to me what is the best solution:
   -> defaulting to `user` (currently coded as BB) with a printed warning
   -> defaulting to `BB` with a printed warning
   -> making it crash altogether rather than the above 2 solutions with warnings that might be missed
   
Thoughts ?

thjsal:
Hi @sguillot! I think what you now coded looks good to me (so apparently the second option of your list). I think making it crash might be even better (and it actually should already crash based on the checks made even before calling these functions), but as far as I remember, doing it in a clean way in Cython was not very trivial, so I would leave it as future enhancement if needed.

sguillot:
Ok, so we can open an issue for future improvements, and we are good to go with this PR.

sguillot (approved):

PR #308, opened 2023-05-20
Solving numerical problems with zero counts - X-PSI v2.0.2.

thjsal:
Solutions to solve numerical problems with zero counts in likelihood calculation. Still to be tested, possibly improved, and documented.


thjsal:
I made small updates to these fixes and re-checked that they work fine (giving reasonable likelihoods instead something close minus infinity) in my example with the data where some channels have zero counts. And I tested that no effects seem to be for the usual examples. I documented the changes in the ChangeLog and changed the version number, and I think it could be ready for a merge unless people still want to test this.

yveskini (approved):

PR #322, opened 2023-08-17
requirements for cython version

sguillot:
in environment yml files, and in installation documentation.

### DIFF
--- a/basic_environment.yml
+++ b/basic_environment.yml
@@ -3,7 +3,7 @@ channels:
   - defaults
 dependencies:
   - numpy
-  - cython
+  - cython ~= 0.29
   - matplotlib
   - scipy
   - wrapt
--- a/docs/source/HPC_systems.rst
+++ b/docs/source/HPC_systems.rst
@@ -236,7 +236,7 @@ Let's start by loading the necessary modules and creating a conda environment. A
    conda create -n xpsi_py3 python=3.10.6
    conda activate xpsi_py3
    conda install -c conda-forge mpi4py
-   conda install cython
+   conda install cython ~= 0.29
    conda install scipy
    conda install matplotlib
    conda install wrapt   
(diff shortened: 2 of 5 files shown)


sguillot:
What I haven't tested is to create an environments with `cython = 3.0` and try to install X-PSI to double-check if the requirement in the `setup.py` catches it.

thjsal (approved):
Without actually testing it, it looks ok to me and I think it could be merged.

thjsal:
I now tested if the `setup.py` catches the cython requirement (if creating environment with wrong cython version), and it seems to me that it does not, since requirement catching (in general) is not working for me. I get this warning:

```
Cython.__version__ == 3.0.0
Using Cython to build extension from .pyx files...
/home/tuomo/anaconda2/envs/xpsi_py3_test/lib/python3.11/site-packages/setuptools/__init__.py:84: _DeprecatedInstaller: setuptools.installer and fetch_build_eggs are deprecated.
!!

        ********************************************************************************
        Requirements should be satisfied by a PEP 517 installer.
        If you are using pip, you can try `pip install --use-pep517`.
        ********************************************************************************

!!
  dist.fetch_build_eggs(dist.setup_requires)
```
Upgrading setuptools using `pip install --upgrade setuptools` or `pip install --use-pep517 --upgrade setuptools` did not help.

thjsal:
Maybe this is because installing with setup.py is anyway a deprecated method? I suggest we could still merge this since the correct Cython version is now already there when creating the conda environment. But I see now that installation instructions for HPC systems need to still be updated when not using `basic_environment.yml` to make sure not using too new Cython version. I can do that soon.

Anyway, I removed the closing of the linked issue, since this is only a temporary solution (X-PSI should be preferably fixed to work with Cython3), and versions for Numpy, etc., as mentioned in the issue, should probably be specified as well.

thjsal:
Ah, I see now that you actually already fixed the documentation for HPC installation as well. So it looks good to me then.

PR #326, opened 2023-09-08
Update Surface_radiation_field_tools.ipynb

sguillot:
Here's a PR for just one extra `#`


thjsal (approved):
I compiled it, and it works!

PR #329, opened 2023-09-08
Support for X-ray polarimetry

thjsal:
Calculation of polarized X-ray pulses (with Stokes parameters I, Q, and U) implemented. An example and tutorial has also been drafted. The code should still be optimized, and better documented, and the statistical analysis part should be checked/finished.

Currently, there are separate integrators for Stokes parameter calculation. It might be better to merge them with the normal integrators to avoid repeating similar codes very many times. But need to think if that can be done so that extra computation is avoided when polarized signals are not needed.


drannawatts (approved):
Go for it!!

CHANGELOG.rst
Basdorsman, line 1: Technically it's first 5D interpolations to interpolate 3 parameters (creating a 2D dataset) and then 2D interpolations for the remaining 2 parameters.

Basdorsman (approved):
Changelog looks fine. I tested the 1 spot example and it works fine.

PR #331, opened 2023-09-26
Numerical fix in KL-divergence calculation

thjsal:
Preventing the prior probability values from being exactly zero (or negative) for KL-divergence calculation to prevent infinite values for the log-probabilities.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -34,6 +34,20 @@ and this project adheres to
 .. ^^^^^^^^^^^
 
 
+[pre-v2.1.1] - 2023-09-26
+~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Fixed
+^^^^^
+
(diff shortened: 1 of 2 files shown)


thjsal:
Yes, I think it is the 1D-probability value. But shouldn't flat prior still have with non-zero probability values? Unless prior probablility is zero everywhere? I doubt we should have priors that are zero everywhere.

thjsal:
If prior would be zero (or negative) to all `x` values, it would also mean that we get "inf" KL-divergence values (since `np.log(prior)=inf `if `prior=0`) every time that happens, but I have not seen that happening for flat priors.

DevarshiChoudhury:
Yeah, that's true. I haven't seen it either, and also generally thinking about it, having a prior that's flat everywhere should mess with evidence calculation as well, so I guess that it's also an indication for something going wrong.

Alright, that was my only query. I will merge it then.

DevarshiChoudhury:
Um, not sure why the branch didn't merge and only the issue closed

DevarshiChoudhury (approved):

PR #334, opened 2023-10-05
Fixing module generator

thjsal:
Problems with module generators fixed so far, except for the CustomInstument not being yet compatible with the example instrument files.


thjsal:
I fixed the CustomInstrument in the example just by adding the working version for the tutorial to the module-generator as commented lines, and then instructed in the tutorial to comment/uncomment some lines to make the modules work (but in the example modules they are already fixed, so one can directly run them now).

DevarshiChoudhury (commented):
All seems good! Thanks @thjsal! 🙂

DevarshiChoudhury (approved):

PR #337, opened 2023-10-30
Credible intervals for multiple models in corner plots

thjsal:
Displaying and saving the credible intervals for all plotted posteriors fixed to work for multiple models and not only for multiple runs.


yveskini (approved):

sguillot:
So this is for multiple models only right?  Not for multiple runs with the same model ?

yveskini:
@sguillot , It works for both now!

thjsal:
@sguillot As Yves said, it works for both. Previously, it worked only for multiple runs. After these changes, it works for multiple models too. However, if having multiple models and multiple runs at the same time, it might not work as one would expect, since X-PSI forces the multiple runs to be then combined for each model. So it is (as before) not possible to plot multiple runs and multiple models separately at the same time.

sguillot:
Ah ok, so if it works for multiple runs, I suggest to add an example in the notebook.

I tried setting `show_vband=0` in the corner plot of `IDs=OrderedDict([('ST', ['run_seed0','run_seed42',]),]),` and the corner plot still showed 1 vertical band. That's not how it is supposed to work ?

thjsal:
It only works if setting also `credible_interval_1d_all_show=True`. Some examples are already there (see e.g. block 13 in https://xpsi-group.github.io/xpsi/Post-processing.html), but those could be sure commented more clearly.

sguillot:
Thanks for the clarification. I did managed to make it work by adding `credible_interval_1d_all_show=True` in the corner plot with two runs. But what I suggested is to show an example with the `show_vband=2` for example (or `show_vband=0`) in the notebook. I don't see it at the moment in the committed version of this PR.

Another comment is that I find it weird to force `credible_interval_1d_all_show=True` when we want to show none of the vertical bands. For example, if I want to compare "many" runs, like I did for Athena, and not show the vertical band, the corner plot will show the credible intervals values of the multiple runs. See below:
figure

thjsal:
Thanks, I see your point now! I'll check how to make `show_vband` working without need for using `credible_interval_1d_all_show=True`.

sguillot:
I opened issue #339 that could be fixed with this PR. Even though the issue existed before the PR.

thjsal:
I fixed the code for `show_vband` and for the crash if setting `annotate_credible_interval=False`. I did some additional fixes to make sure the new features work also when `bootstrap_estimators=True`.

I can add examples to the Post-Processing notebook once the branch with xtick and ytick changes to signals plots is merged to this.

thjsal:
Examples of the new features (including `show_vband`) have now been added to the Postprocessing tutorial. Instead of adding even more corner plots there, I combined 3 of them to just directly show more features in one plot (avoiding repeating the large corner plot very many times). 

Please let me know if it seems ok to you @sguillot @yveskini.

sguillot:
> Please let me know if it seems ok to you @sguillot @yveskini.

That looks good! (I didn't re-run the notebook to double-check though)

sguillot (commented):
I'm not sure what your modification does, but it broke the post-processing notebook:
Screenshot 2023-11-10 at 10 10 37

yveskini:
@sguillot, we changed the show_vband to be a list. show_vband=None if one does not want to show any of them or show_vband=[0,1] if one wants to show it for the first and second plots. @thjsal is planning to update the notebook sometime before it gets merged.

sguillot:
I see. Indeed, it makes more sense to list individual bands rather than provide the number.
It works well :-)
I'll commit the notebook with the update on show_vband

thjsal:
Thanks @sguillot ! I just updated the documentation to account for the newest changes. One thing to note is that @yveskini 's updates changed also the default behaviour of the corner plotting so that it now calculates the 1D credible intervals for all the runs, even when not showing all of them. Potentially that could increase the time needed to produce a plot and not wanting to show/save all the credible intervals. However, in the basic examples, the difference in time is not easily noticed, so hopefully that is not a problem with much larger number of runs either.

I am thinking of merging this now, unless someone has still comments/objections.

sguillot:
> Potentially that could increase the time needed to produce a plot when not wanting to show/save all the credible intervals. However, in the basic examples, the difference in time is not easily noticed, so hopefully that is not a problem with much larger number of runs either.

I'll let you know when I try to plot the results for 10 NewAthena runs :-)

> I am thinking of merging this now, unless someone has still comments/objections.

Ok for me!

Issue #243, opened 2023-02-01
Add extra y-ticks options for the 2D pulse plots in the post-processing 


sguillot:
In a manner similar to what was done for the notebooks. See PR #240

PR #338, opened 2023-11-01
Extra ticks options for the 2d pulse plots in the post processing

thjsal:
Extra ticks options added as suggested in the issue #243 . The post-processing tutorial has been updated to show how to use these options.

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -41,6 +41,7 @@ jobs:
     - name: install gsl
       run: |
         sudo apt-get install -y libgsl-dev
+    - name: Make sure metadata not too new for flake8
       run: |
         pip install "importlib-metadata<5.0"
     - name: Lint with flake8
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -31,6 +31,9 @@ jobs:
         cache-env-key: env-envkey-${{ github.sha }}-${{ github.run_attempt }}
         environment-file: basic_environment.yml
         environment-name: xpsipy3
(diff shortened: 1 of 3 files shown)


docs/source/Post-processing.ipynb
sguillot, line 1: In the notebook, for the 1st residual plot (cell 21 I think), I suggest explaining that the list of yticks are given for the 3 plots, alternating the main plot and the color bar y-ticks. A comment directly in the cell is sufficient. I know it is in the docstring of `_signalplot` but that will help the user understand I think.

And for the signal plots, you give an example with the x-ticks, which are not alternating like the y-ticks. This is a bit confusing. Maybe just add a small comment that these 3 `None` should not be changed to not have x-ticks on the color bar.

thjsal: One thing that makes this more confusing is that for residual plot the order of subplots is (have not checked why): 
`top plot -> top color bar -> middle plot -> middle color bar -> bottom plot -> bottom color bar
`but for the other signal plots the order is
`top plot  -> middle plot -> bottom plot -> top color bar -> middle color bar -> bottom color bar
`
I'll add more comments about these in the tutorial.

thjsal: Comments are added now to the Tutorial. CI-tests do not seem to pass anymore, but that seems to be a different issue that I just reported.

thjsal: CI-tests are working again, since the issue the issue with them corrected and merged to this branch....

Anyway, I guess it is probably fine to merge this branch now to the other post-processing update branch, so I can continue updating the PostProcessing tutorial notebook there without getting conflicts between the branches.

Issue #340, opened 2023-11-04
GitHub actions fail due to flake8 not working anymore


thjsal:
Seems that our GitHub Actions (continuous integration -tests in particular) are now failing because flake8 used for pytesting is not working anymore perhaps due to an inconsistent version of `importlib-metadata` package (error message is similar to that one here: https://github.com/PyCQA/flake8/issues/1701):

```
Run # stop the build if there are Python syntax errors or undefined names
Traceback (most recent call last):
  File "/home/runner/micromamba-root/envs/xpsipy3/bin/flake8", line 10, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/runner/micromamba-root/envs/xpsipy3/lib/python3.[12](https://github.com/xpsi-group/xpsi/actions/runs/6755894386/job/18364788186#step:7:13)/site-packages/flake8/main/cli.py", line 22, in main
    app.run(argv)
  File "/home/runner/micromamba-root/envs/xpsipy3/lib/python3.12/site-packages/flake8/main/application.py", line 375, in run
    self._run(argv)
  File "/home/runner/micromamba-root/envs/xpsipy3/lib/python3.12/site-packages/flake8/main/application.py", line 363, in _run
    self.initialize(argv)
  File "/home/runner/micromamba-root/envs/xpsipy3/lib/python3.12/site-packages/flake8/main/application.py", line 343, in initialize
    self.find_plugins(config_finder)
  File "/home/runner/micromamba-root/envs/xpsipy3/lib/python3.12/site-packages/flake8/main/application.py", line [15](https://github.com/xpsi-group/xpsi/actions/runs/6755894386/job/18364788186#step:7:16)7, in find_plugins
    self.check_plugins = plugin_manager.Checkers(local_plugins.extension)
                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/runner/micromamba-root/envs/xpsipy3/lib/python3.12/site-packages/flake8/plugins/manager.py", line 363, in __init__
    self.manager = PluginManager(
                   ^^^^^^^^^^^^^^
  File "/home/runner/micromamba-root/envs/xpsipy3/lib/python3.12/site-packages/flake8/plugins/manager.py", line [24](https://github.com/xpsi-group/xpsi/actions/runs/6755894386/job/18364788186#step:7:25)3, in __init__
    self._load_entrypoint_plugins()
[...]

PR #341, opened 2023-11-05
Fix to failing continuous integration GitHub Actions

thjsal:
Fixed here the `flake8` issue by correcting python version setting when creating conda environment with micromamba. Apparently, we have so far in practice always used the newest python version for all the CI-tests, even though trying to do it separately for 3.8, 3.9, 3.10, and 3.11. And now when the newest version is 3.12, all the tests failed because the `flake8 `package seems broken for that version (at least I could not install it without getting an inconsistent version for the needed `importlib-metadata` package as mentioned in the issue #340). 

I fixed the conda environment creation to actually use the intended python versions and also switched from the deprecated `provision-with-micromamba` method to the `setup-micromamba` method as explained [here](https://github.com/mamba-org/provision-with-micromamba#migration-to-setup-micromamba%60).

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -25,15 +25,18 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: Install Conda environment with Micromamba
-      uses: mamba-org/provision-with-micromamba@main
+      uses: mamba-org/setup-micromamba@v1
       with:
-        cache-env: true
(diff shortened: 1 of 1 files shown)


thjsal:
Of course, this solution won't help if we want to add `python3.12` version to our continuous integration tests. In that case, we could perhaps just skip the linting with flake8.

drannawatts (approved):

Issue #306, opened 2023-04-20
No API for CornerPlotter showing on documentation page


sguillot:
There are some examples of `CornerPlotter` in the tutorials, but the full API does not show on this page: https://xpsi-group.github.io/xpsi/postprocessing.html

thjsal:
I think that page mostly shows the docstrings for all the CornerPlotter methods. Is there something specific that you find missing there?

Or you were expecting to see more examples of how to use the methods, as in here: https://xpsi-group.github.io/xpsi/Post-processing.html?

sguillot:
```
I think that page mostly shows the docstrings for all the CornerPlotter methods. 
```
That's the thing, it doesn't show the docstring of CornerPlotter, only for `SignalPlotter`, `ResidualPlotter`, `PulsePlot`, `SpectrumPlot`

thjsal:
I see you are right! Need to check why that is so.

thjsal:
The Cornerplotter API actually shows up when I compile the documentation locally on my laptop. So for some reason, GitHub does not compile it in the same way.

PR #346, opened 2023-11-10
CornerPlotter broken API fixed for web pages

thjsal:
As mentioned in issue #306 the CornerPlotter API has not been showing up in the documentation pages for a while. That is now fixed by installing getdist and nestcheck in the GitHub Action that builds the documentation pages. Apparently, those packages were required for CornerPlotter to appear in the documentation. 

I only fixed this for the Python3 web pages. For Python2 pages the installation of getdist and nestcheck did not seem to work with same command, and I did not spend time checking alternative ways to do that. Having the full API only in the Python3 pages is probably still ok.


thjsal:
I already manually ran the documentation building and page deployment from the new branch, so, therefore, the web pages already show the CornerPlotter API correctly. Only thing that I don't know is why docstrings only for some functions are shown, but not for all of them. But that is maybe determined somewhere.

drannawatts (approved):
Approved....

PR #348, opened 2023-11-12
Module generator bug fixed to correctly import Custom files

DevarshiChoudhury:


thjsal:
The problem is that adding this dot in front of the imported module breaks modules if running them directly. Like this:
```
(xpsi_py3) tuomo@tuomo-XPS-13-9310:~/xpsi/xpsi_group/xpsi/examples/examples_module_generator$ python _auto_modules/main.py @config.ini 
/=============================================\
| X-PSI: X-ray Pulse Simulation and Inference |
|---------------------------------------------|
|                Version: 2.1.1               |
|---------------------------------------------|
|      https://xpsi-group.github.io/xpsi      |
\=============================================/

Imported GetDist version: 1.4.5
Imported nestcheck version: 0.2.1
Parsing configuration file...
--main-import-statements
...
Configuration file parsed.
Rank reporting: 0
  File "/home/tuomo/xpsi/xpsi_group/xpsi/examples/examples_module_generator/_auto_modules/main.py", line 858, in <module>
    from .CustomInstrument import CustomInstrument
ImportError: attempted relative import with no known parent package
```

DevarshiChoudhury:
Let me test if I face the same issue. I don't think I have directly run the `main` module generated using py3 module generator. I was facing import issues when importing the `main` module from a notebook. If I am also able to reproduce the error above, I suppose a good way would to be put a `if __name__ == '__main__' condition when importing the `Custom` modules in `main`

DevarshiChoudhury:
@thjsal I have pushed another commit to account for relative imports.

xpsi/module_generator.py
```diff
@@ -1433,14 +1446,20 @@ def __call__(self, parser, namespace, values, option_string=None):
...
+    from .{0} import CustomBackground
     '''.format(args.custom_background_module)
     )
 elif args.background_shared_class:
     for _instrument in args.instruments:
         module += (
         '''
-from {0} import {1}_CustomBackground
+if __name__ == '__main__':
+    from .{0} import {1}_CustomBackground
```
thjsal, line 1460: Probably no dot should be here, right?

thjsal: And actually, it probably should be:
```
if __name__ == '__main__':
  from {0} import {1}_CustomBackground
          '''.format(args.custom_background_module,
                     _instrument)
else:
  from .{0} import {1}_CustomBackground
          '''.format(args.custom_background_module,
                     _instrument)
```

DevarshiChoudhury: Ah yes that's correct. The `else` condition is fine in the commit. The `if` condition has the typo that you caught

thjsal: I pushed a commit to fix that.

thjsal (approved):

Issue #336, opened 2023-10-13
Set maximum version of the NumPy dependency (NumPy 2.0 is coming)


evertrol:
I [previously wrote](https://github.com/xpsi-group/xpsi/issues/319):

> Aside, something similar should be the case for the NumPy dependency. There should be a minimum version requirement, and a maximum version requirement, for any dependency.

Apparently, I'm slightly clairvoyant, because possibly somewhere end of this year, [NumPy 2.0 will be released](https://github.com/numpy/numpy/issues/24300).

I don't expect the normal API will actually change. But under the hood, a lot might change. And there is some  use of NumPy through Cython in XPSI, which may cause unforeseen problems. Patches will likely come for Cython to account for that, but that will be in Cython 3, not 0.29.

So for the moment, you might want to limit the NumPy version to 1.x (`~= 1.21` or whatever is a good minimum).

----

Also, spread the word, because this may cause plenty of problems for people trying to install a random half-maintained quick-thrown-together Python package(*) that depends on NumPy. The linked NumPy 2.0 issue/announcement states solutions, but these are for package maintainers, not for innocent users.

(*) I mean one of those many "scientific" Python packages out there that people use in some random project.

Issue #335, opened 2023-10-12
Installation instruction updates


thjsal:
Things to be fixed/updated in the installation instructions:

- MultiNest should be installed before installing mpi4py using `conda install mpi4py`, or the path to `mpif90` compiler should be specified to be not the apparently broken`mpif90` created by mpi4py installation.
- GSL can be also installed with `conda install gsl`, and in some cases (especially for newest Mac) it seems even required to get gsl linked correctly with X-PSI. Conda installing gsl seems to work in Linux-machines too. So we can probably advice installation of GSL to be always using conda.
- The version of Sphinx needs to be specified to be below 7.0.0, since otherwise compiling the documentation causes a `"'style' is undefined"` error as in https://github.com/kivy/kivy/issues/8230. Alternatively, the notebooks (apparently `Emitting_patterns_2Dprojection.ipynb` at least) should be updated to be compatible with the latest sphinx version.

thjsal:
We should also add `sudo apt install mpich` to instruction, to make sure `mpif90` is installed.

PR #352, opened 2023-11-25
Installation instructions updated

thjsal:
Installation instructions are updated based on the latest feedback and issues #335 #336 .

### DIFF
--- a/basic_environment.yml
+++ b/basic_environment.yml
@@ -2,7 +2,7 @@ name: xpsi_py3
 channels:
   - defaults
 dependencies:
-  - numpy
+  - numpy < 2.0.0
   - cython ~= 0.29
   - matplotlib
   - scipy
--- a/docs/source/install.rst
+++ b/docs/source/install.rst
@@ -211,6 +211,12 @@ compilers (e.g. ``gcc`` and ``gfortran``), ``cmake``, ``blas``, ``lapack``, and
 
     sudo apt-get install cmake libblas-dev liblapack-dev libatlas-base-dev
 
+To have MPI-wrapped compilers, one should also install ``mpich`` if not installed already:
+
+.. code-block:: bash
+
+    sudo apt install mpich
+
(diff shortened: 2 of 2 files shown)


MHoogkamer (approved):

MHoogkamer (commented):
`xpsi/include/rayXpanda/*.o` does not work for me because I never installed rayXpanda (I get: `rm: cannot remove 'xpsi/include/rayXpanda/*.o': No such file or directory`). Not sure why this is needed?

thjsal:
> `xpsi/include/rayXpanda/*.o` does not work for me because I never installed rayXpanda (I get: `rm: cannot remove 'xpsi/include/rayXpanda/*.o': No such file or directory`). Not sure why this is needed?

I followed what @evertrol  suggested here:
https://github.com/xpsi-group/xpsi/issues/335#issuecomment-1761542722

I am not sure if that is actually needed, since we don't usually use rayXpanda. Interesting though that you don't have any *.o files there, since I seem to have them there even in systems where I have not installed rayXpanda.

evertrol:
Yes, it's a bit confusing if there are no actual files to remove. Perhaps it's an idea to add the `-f` option to `rm` (together with `-r` where needed), so that `rm` will just ignore non-existent files and directory and not complain about it and fail.

evertrol:
> I am not sure if that is actually needed, since we don't usually use rayXpanda. Interesting though that you don't have any *.o files there, since I seem to have them there even in systems where I have not installed rayXpanda.

Perhaps I'm overlooking a few things (after all, this is not software I often, if at all, use), but in `setup.py`, aren't the `inversion.o` and `deflection.o` created by default when `rayXpanda` can't be imported? As some sort of dummy interface? (Even if the `.so` shared library files may never be created from the object files.)

Though that does not agree with MHoogkamer's assessment that `rayXpanda` was never installed, but the `.o` files are also not present. 🤔

PR #356, opened 2023-12-19
Updates to publications list

drannawatts:
Added links to Salmi et al. 2023, Vinciguerra et al. 2023, 2024, Kini et al. 2024.

### DIFF
--- a/docs/source/applications.rst
+++ b/docs/source/applications.rst
@@ -1,6 +1,6 @@
 .. _applications:
 
-Applications
+Publications
 ------------
 
 X-PSI has been applied in the following studies. In addition to giving an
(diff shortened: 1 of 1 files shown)


yveskini:
Shouldn't it be  Kini et al. 2023b instead of  Kini et al. 2024?

drannawatts:
Actually no! If you check the MNRAS website carefully you will see that it is going to appear in the first volume of 2024

thjsal:
It looks good to me. A small inconsistency (I think not important to fix) is that Kini 2024 is linked to the published paper, but all the other papers are linked to the corresponding ADS pages.

thjsal (approved):

thjsal:
There is also this paper from Afle: https://arxiv.org/abs/2304.01035 (but that could be added later as well if needed)

drannawatts:
Yes, right now the ADS link is to the tmp.MNRAS file (which I think will change) so I figured I'd link to the actual paper for now.  When I go in to update the BRAVO paper to the published version I can hopefully update this too (and indeed add Afle, good point!).

PR #362, opened 2024-01-24
Update synthesise.pyx

Basdorsman:
clarified docstring

### DIFF
--- a/xpsi/tools/synthesise.pyx
+++ b/xpsi/tools/synthesise.pyx
@@ -61,7 +61,8 @@ def synthesise_exposure(double exposure_time,
     :param tuple component_phases:
         For each component, a C-contiguous :class:`numpy.ndarray` of phases
         in cycles at which the model :obj:`signal` is evaluated on
-        the interval ``[0,1]``.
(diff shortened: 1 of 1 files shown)


Basdorsman:
I wanted to close issue 361 with a pull request but I think I failed

xpsi/tools/synthesise.pyx
```diff
@@ -61,7 +61,8 @@ def synthesise_exposure(double exposure_time,
     :param tuple component_phases:
         For each component, a C-contiguous :class:`numpy.ndarray` of phases
         in cycles at which the model :obj:`signal` is evaluated on
-        the interval ``[0,1]``.
+        the interval ``[0,1]``. Typically this is more finely spaced than the
+        to be synthesised data.
```
thjsal, line 65: Would it be more clear if saying "the resulting synthesised data" instead of "the to be synthesised data"?

Same comment also fir below, where this text is repeated.

thjsal:
I made the change, and merging to main seems ok to me.

thjsal (approved):

PR #365, opened 2024-02-05
matched vectors and lowered num energies for faster runs

Basdorsman:


Basdorsman:
Test run takes 24 minutes on 8 cores

thjsal:
Perhaps it just becomes slower if using that many cores? At least for me the sampling takes only ~ 1.6 minutes (even with 128 energies) when using one core.

thjsal:
And dropping the num_energies to 64 brings the sampling time to ~ 1.2 minutes for me. Anyway, these changes are fine to me.

thjsal (approved):

PR #370, opened 2024-02-27
PDT-U projection tool bug fixed and units set to roman font.

thjsal:
Bug in the projection tool for PDT-U model fixed (the components of the secondary region were not shown). In addition, the font type of the units was changed from italic to roman. Here a figure before the update
Projection_from_PrimarySuper_old
and here the same figure after the update:
Projection_from_PrimarySuper_new

I also checked that the code still works correctly for ST-U.

### DIFF
--- a/xpsi/utilities/ProjectionTool.py
+++ b/xpsi/utilities/ProjectionTool.py
@@ -307,6 +307,7 @@ def plot_projection_general(dictVp, model, POV = "", ThetaDisplay = "",antiphase
             elif ('-U' in model):
                 ind = model.find('-')
                 hot_name = model[:ind]
+                hot_name_s = model[:ind]
                 hotspot_check(hot_name,DICT_VECTOR,'p')
                 hotspot_check(hot_name,DICT_VECTOR,'s')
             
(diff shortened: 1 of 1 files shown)


thjsal:
As seen from the figures, the color choices are also not very optimal for PDT-U. It is hard to separate the circles, since the color looks the same if the primary circle is in the front side and if the secondary circle is on the back side of the star. I think we would need to either change the colors or make curves dashed instead of dimmed when they are behind the star.

DevarshiChoudhury:
Do you have an example figure that you can paste here, of where it doesn't seem to work @thjsal ?

DevarshiChoudhury:
I think colour changes might be better since having more dashed lines would make the plot look too busy. 
Suggestion:
Primary colours: Red and blue
Secondary colours: Purple and goldenrod

thjsal:
> Do you have an example figure that you can paste here, of where it doesn't seem to work @thjsal ?

The first figure I posted is the example figure that does not work. You see that it does not plot all the circles it should. The second figure is the same figure after the change in the code.

thjsal:
Probably then need to look at the colors a bit later (perhaps when Serena is back). How they are set in the code seems a bit complicated to me and changing the colors may also break the current connection between colors and temperatures of the spots.

serenavinciguerra89:
Everything looks fine to me :)

serenavinciguerra89 (approved):

PR #372, opened 2024-03-10
The default imaging extension changed

thjsal:
The default imaging extension has been changed from ``uniform.pyx`` to ``PDT_U.pyx`` to avoid the need for X-PSI re-installation when computing neutron star images for different nested models. The tutorials have been updated to account for this change where necessary.

Some extensions (like the r-mode extension) are still only possible to use if re-installing X-PSI after changing the content of the `local_variables.pyx` file.


thjsal:
As a part of this pull request, this is the tutorial that is going to be back as a page in the X-PSI documentation (in case @drannawatts wants to check):

https://github.com/xpsi-group/xpsi/blob/358-change-default-extension-for-image-plane/docs/source/Global_surface_emission.ipynb

It is very technical, but I think it could be useful (it also used to be there before X-PSI version 1.0).

Basdorsman (approved):
Approving with the note that it would be even better if we could simultaneously use the various local_variables handlers such as PDT_U, ring and r-modes. Typically, we don't need them simultaneously (in the global surface emission tutorial you need to switch mid tutorial), but the software would be more flexible if you didn't have to reinstall everytime you wanted to switch. This can be implemented once it becomes relevant for someone.

PR #374, opened 2024-03-23
Example script updates

thjsal:
X-PSI example scripts udpated to allow same compactness as commonly allowed for NICER analyses and the commonly used multiple imaging restrictions added.

In addition, a comment added to Synthesise notebook on how to save synthetic data without noise if interested (rounding the numbers to nearest integer in this example).


Basdorsman:
My code has 1.45 instead of 1.505, I don't know what's the best number, but at least they're closer to each other. Please advise :)

Basdorsman (commented):
so 1.505 is a better compactness limit? What about 1.45? My code is currently using 1.45. Is that too low?

Basdorsman (approved):
Approving with the note that both compactness limits I mentioned are fine, 1.7 was just too much

PR #376, opened 2024-04-22
Gap setting added for texts above corner plot diagonal.

thjsal:
Added a setting for modifying the gap between the texts printed above the corner plot diagonal to avoid the different reported credible intervals overlapping with each other when increasing the font size for large corner plots.


thjsal:
Here an example plot with default settings:

download

Here an example plot with modified settings (`ci_gap = 0.17` instead of `ci_gap = 0.12`):

image

drannawatts (approved):
Looks good to me, merge away!

PR #379, opened 2024-04-26
`module_generator` bugfix to choose correct atmosphere extension.

DevarshiChoudhury:


xpsi/module_generator.py
```diff
@@ -1930,8 +1930,10 @@ def get_member_settings(model):
 get_member_settings(args.hot_region_model[0])
 get_bounds_and_values(args.prefix[0], args.hot_region_model[0])
 
-module += (
-'''
+if 'NSX' in args.hot_atmosphere_model or 'nsx' in args.hot_atmosphere_model:
+    print(args.hot_atmosphere_model)
```
thjsal, line 1934: Is this printing accidentally left from debugging to here, or purposely?

thjsal:
The fixes look good to me. I also looked at the elsewhere option in module generator (as you mentioned in the original issue), and looks like that no atmosphere selection has been implemented to that yet. So it uses always the default blackbody atmosphere.

DevarshiChoudhury:
@thjsal print statement removed and option for elsewhere added. Tested it and seems to work fine

thjsal (approved):
Looks good to me!

PR #381, opened 2024-04-26
Allow Everywhere for standard Likelihood

thjsal:
Likelihood.py modified so that it can be used for both HotRegion and Everywhere objects without user customization. In case of Everwhere, however, one needs to make a CustomEverywhere where an extra `phase_shift` parameter is added like this (EDIT: THIS IS NOT NEEDED ANYMORE):

```
class CustomEverywhere(xpsi.Everywhere):
    """ Custom surface radiation field globally spanning the surface. """
    
    @classmethod
    def create_parameters(cls, bounds, values={}, *args, **kwargs): 
            ......
             phase_shift = Parameter('phase_shift',
                                      strict_bounds = (-np.infty, np.infty),
                                      bounds = bounds.get('phase_shift',None),
                                      doc = "The phase of the everywhere [cycles]",
                                      symbol = r'$\phi$',
                                      value = values.get('phase_shift',None))
              
              return cls(*args,
                         custom=[...,
                                 phase_shift],
                         **kwargs)
```

This parameter can be fixed to zero when creating an Everywhere instance:

```
values = dict(phase_shift = 0.0)
everywhere = CustomEverywhere.create_parameters(bounds=...,
                                                values=values,
                                                ...)
```

This is to avoid having extra if or try statement in the likelihood evaluation slowing the inference runs (an option disfavored by @yveskini), but still being able to use the same Likelihood syntax for both HotRegions and Everywhere.

We can add phase_shift to be also part of Everywhere class by default, but maybe it is not needed, since for usual applications that class is customized anyway?


thjsal:
I just noticed that for some modules we have used this update produces this error:
`AttributeError: 'CustomSignal' object has no attribute 'surf'`

I don't know why this occurs only for some of the tests (e.g. `TestRun_BB.py` works but `TestRun_PolNum_split_inference.py` does not). But need to still fix this issue.

thjsal:
I found that the issue occured always when having multiple signal objects.I fixed it by changing `signal.surf` to `photosphere.surf `, since it actually makes more sense to save surface model (which is either `photosphere.hot` or `photosphere.everywhere`) to the photosphere class. It is also more consistent with the original code which used always `photosphere.hot`.

thjsal:
Now I also modified the code so that Everywhere class will always create and fix this additional phase_shift parameter to zero without any customization from the user.

thjsal:
I made changes to the coding style following mostly the suggestions of @yveskini. In addition, I moved setting the correct `photosphere.surface` from Likelihood.py to Photosphere.py, since there is no reason why it could not just happen there already.

I tested that the new code also works for the basic examples with HotRegion and Everywhere classes. I also tried to check the do_fast option, but that option seems to be broken even without the changes in this pull request.

yveskini (approved):

PR #384, opened 2024-05-21
A new model parameter and example updates for polarimetry

thjsal:
Added a new spin axis position angle parameter for modeling polarized X-ray. This describes the angle between the observer north and the projection of the pulsar rotation axis on the plane of the sky. The example scripts and the tutorial for polarimetry were also updated.


drannawatts (approved):
That is a lot of commits! 🥇   Clear from me for merging.....  (unless you want to first correct udpated -> updated in the first part of the CHANGELOG!

drannawatts (approved):

PR #386, opened 2024-06-10
Disk

Basdorsman:
Main feature is that the disk is implemented through the signal. Looks like this implementation style makes any new cythonization in disk signal integration is unnecessary (which I did in my fork), since it will be integrated anyways as part of the signal. I am not in a hurry to merge but I would be happy if the code is reviewed and get some feedback. 

I have implemented the slower "split=False" num5D (only for stokes=False), just because I want to compare speeds, but using this option would otherwise never be recommended. 

TestRun_AMXP.py is a relevant example where these features are showcased.


xpsi/cellmesh/integrator_for_azimuthal_invariance.pyx
thjsal, line 1: I guess modifications to this file are breaking the current Github Action likelihood checks, which do not use E_electronrest. Were modifications here actually needed?

Basdorsman: Thank you for finding the probable culprit! 
Personally, I think the cleanest solution would be to clean all num4D, num5D etcetera, hot.pyx's of any unit conversions since they are very low level. Then the low level code would be doing pure unit agnostic interpolation.

The question is then: where do we do the unit conversions (e.g. in energy?) it looks like NSX and Bobrikova atmospheres require different unit conversions, NSX is even using logarithmic energy (see Num4D). 

In the new code, the azimuthal integrator will at least inherit the atmosphere extensions. Perhaps based on that we can enable any necessary unit conversions, just before sending it to eval hot?

thjsal: I guess best would be to have the unit conversions already in the python-files (like HotRegion.py or Photosphere.py)? So people can then customize those based on the atmosphere table used? 

However, I am not sure is that possible. The units are not needed just only for the atmosphere interpolation but also when calculating flux? Or you need to provide 2 different energies as input parameters to each integrator? One for atmosphere interpolation with self-defined units and 1 using the current default X-PSI units?

thjsal: Here is the line I was looking at (where energies are needed besides the atmosphere interpolation):
https://github.com/xpsi-group/xpsi/blob/d39b2e2c84a5591adb043126e28247a33b395f68/xpsi/cellmesh/integrator_for_azimuthal_invariance.pyx#L619

thjsal: In case making all interpolators/integrators unitless seems too much work for this pull request, we could just keep the standard Num4D option unchanged (so compatible with NSX). And then instruct to use a modified X-PSI if want to run tests with Bobrikova Num4D?

Basdorsman: As a side note, one nice implementation I once saw was using astropy.units, where a function may require e.g. flux units, and will then convert as needed inside the function. But indeed, that seems like too much work to me for now.

Basdorsman:
Test now passes. By default the disk is turned off in the photosphere.py so it should not interfere with anyone's work. If it is turned on, it occupies the "custom" option.

thjsal:
@Basdorsman Do you mean the example `CustomPhotosphere.py`? I do not see any changes done to the standard `photosphere.py`, unless you forgot to push them? Or you just mean that the standard photosphere.py has no disk, as before?

Basdorsman:
Yes you're right. It's not in photosphere.py, it is in customPhotosphere.py. And I think we can leave it like that, I think it's fine to use the custom option in this flexible way, and it is not necessary to add the usage in the normal photosphere.py. Maybe that can be done later but I'm sure this is also not the final version of the disk anyways.

thjsal (approved):

thjsal:
I see! I think the new test example (TestRun_AMXP.py) had bug when initializing the hot region object (it was selecting Num4D atmosphere instead of Num5D). I hope it is correct now? At least not all flux is zero anymore.

I tested the old examples too, and all they seem to work still fine. 

I also made version and update and wrote something to the Changelog. Maybe you can extend that if you want. One could also mention the new example script in this page (the others are at least listed there): 
https://github.com/xpsi-group/xpsi/blob/main/docs/source/Example_script_and_modules.rst

But otherwise, I think it could be merged.

drannawatts (approved):

PR #388, opened 2024-06-24
UltraNest

MHoogkamer:
UltraNest as an additional sampler


xpsi/Sample.py
```diff
@@ -65,8 +69,6 @@ def ensemble(likelihood, prior, MPI = True, **kwargs):
                 pool.wait()
                 _sys.exit(0)
 
-            from xpsi.EnsembleSampler import EnsembleSampler
```
thjsal, line 68: I guess removal of emcee imports are breaking the GitHub Action tests. Did you remove these by purpose?

thjsal: Well, at least the error messages are complaining about emcee:
https://github.com/xpsi-group/xpsi/actions/runs/9643508619/job/26593654031

.gitignore
thjsal, line 1: Maybe this file should not be removed?

xpsi/Sample.py
```diff
@@ -1,9 +1,14 @@
-__all__ = ["nested", "ensemble"]
+__all__ = ["nested", "ensemble", "ultranested"]
 
 from xpsi.global_imports import *
 from xpsi import _comm, _rank, _size
 from xpsi.utils import make_verbose
 from xpsi import Likelihood
+from xpsi.Posterior import Posterior
+
+from xpsi.EnsembleSampler import EnsembleSampler
```
thjsal, line 9: Ah, I see now the problem is that you changed the location of the emcee import such that now emcee is required to be installed, instead of that being only needed if actually using it.

xpsi/Sample.py
thjsal: Okay, I found the problem is actually the re-location of the emcee import that I mentioned below.

xpsi/Sample.py
thjsal: Your new pull request seems to solve this issue.

.gitignore
MHoogkamer: Agreed, this wasn't on purpose. It's now restored!

thjsal: That looks good now!

thjsal:
The pull request looks good to me. I checked that it at least did not break the existing MultiNest and emcee sampling examples. Probably at some point it would be good to have an UltraNest example as well. 

One thing I modified was still the try-except imports of the samplers, so that you don't need to have all the samplers installed if you just want to use one of them.

I suggest to fill some text to the Changelog (including the date), and then merge. I already updated the version numbers in those 3 files where needed.

When merging, you can do "squash and merge" if just want to have one clean commit to the main branch. Unless you think having all the small commits there is useful.

thjsal (approved):

PR #394, opened 2024-07-20
Remove unused modules

AlexKurek:


drannawatts:
Hi @AlexKurek sorry to take so long to get back to you on this. There are lot of modifications here, and we want to review these carefully. We have a hack week planned in December and will probably take a look then. Can you give us a bit more information about the general aim of these changes though?

AlexKurek:
@drannawatts The goal is just to clean up the code. I learned some linters like Russ and Pylance using astrophysics related codes.

PR #395, opened 2024-08-12
Updating PSR J0437 publication link

drannawatts:
Linked to published version of J0437 paper.


thjsal:
Looks otherwise fine to me, but using a link directly to ApJL for J0437 seems a bit inconsistent to what is used for all the other papers (where links go always to the ADS pages).

thjsal:
I think ADS updates those once per week, so this may happen in a few days. But sure the link can be changed later again.

thjsal (approved):

PR #402, opened 2024-11-07
Default attenuation column corrected in module_generator

### LINKED ISSUE #397: Wrong default column for interstellar attenuation file in module_generator
Module generator uses column 1 (indexed from 0) as default to extract interstellar attenuation from file. But the usual table for interstellar absorption `interstellar_phot_frac.txt` has this value in column 2. 
This is not much, but should we change this since it can be quite misleading and easy to make a mistake there ?

lmauviard:
Simple change in the module generator so the default attenuation column matches the one from the TBabs file that has always been used for now `interstellar_phot_frac.txt`.
This could be quite misleading otherwise since there is no error raised and this changes greatly the inferred radius. 

Closes #397

### DIFF
--- a/xpsi/module_generator.py
+++ b/xpsi/module_generator.py
@@ -829,7 +829,7 @@ parser.add_argument('--attenuation-path', type=str, help='{0} attenuation file.'
 parser.add_argument('--attenuation-energy-column', type=int, default=0,
                     help='Column (zero-indexed) containing the energies in the attenuation file.',
                     comment=True)
-parser.add_argument('--attenuation-column', type=int, default=1,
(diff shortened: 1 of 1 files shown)


thjsal:
Looks good to me, although I suggest also to update the example that is supposed to generated using the module generator, i.e. this line:

https://github.com/xpsi-group/xpsi/blob/a3fd654ca4658825253b46d563d571e766b2f855/examples/examples_module_generator/_auto_modules/CustomInterstellar.py#L54

lmauviard:
I did the change you mentionned !

thjsal (approved):

PR #409, opened 2024-11-19
Fix build_docs and JOSS actions

drannawatts:
I think this mimics the changes that were made to the NEoST actions a few weeks ago - but the actions have other differences so possible we break something else.


nr1118 (commented):
These changes look good to me and it seems that the checks have passed as well

nr1118 (approved):

PR #414, opened 2024-11-29
Avoid taking log of 0 when channel 0 is included.

thjsal:
The lower limit for the channels in signal plots was set to 0.001 instead of 0, to avoid only showing the channel 0 when it is included and logarithmic scale is used. 

However, I see now that the original behaviour of the code was not in principle wrong. Using a logarithmic scale is just a bad choice if including channel 0 (it makes zeroth bin infinitely large). Do you think this change is still worth doing? 

Attached is an example of the pulse plot including channel 0, using the old code  (only the counts from channel 0 are visible the middle bottom panel).
pulse_old
and here using the new code:
pulse_new

### DIFF
--- a/xpsi/PostProcessing/_pulse.py
+++ b/xpsi/PostProcessing/_pulse.py
@@ -449,7 +449,7 @@ class PulsePlot(SignalPlot):
                                        rasterized = self._rasterized)
 
         incident.set_edgecolor('face')
-        self._ax.set_ylim([ref.energy_edges[0],
+        self._ax.set_ylim([_np.max([ref.energy_edges[0],0.001]),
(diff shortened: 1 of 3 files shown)


thjsal:
So I think possible options are:

A) Not to do anything, in which case we can just close this pull request and the corresponding issue.
B) Accept this change as done in this pull request.
C) Modify the lower limit value 0.001 to be something like 0.5, which would make the plot even more "normal looking". And perhaps could modify this change so that it is only applied when logarithmic scale is used.
D) Just print a warning if user tries to make logarithmic plots when channel 0 is included.

drannawatts (approved):
I think this solution is a sensible one!

PR #422, opened 2024-12-02
Remove py2 documentation from build docs action

drannawatts:
Removed what I think are all of the now irrelevant commands from the build_docs action, have also cross-checked with the neost build_docs action.

One question - should we upgrade micromambav1 to v2? But not sure what this would do.

### DIFF
--- a/.github/workflows/build_docs.yml
+++ b/.github/workflows/build_docs.yml
@@ -25,38 +25,6 @@ jobs:
     - name: install gsl
       run: |
         sudo apt-get install -y libgsl-dev
-    - name: check out X-PSI v1.2.1
-      uses: actions/checkout@v3
-      with:
-        ref: python2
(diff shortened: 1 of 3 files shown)


thjsal:
I manually ran the Github build docs using this branch and looks like doc building is now working without errors (see the "Actions").  However, the py3 homepages have still this outdated text in it (including a link that is not working anymore):

image

I suggest to update this as well.

thjsal:
So update this file:
https://github.com/xpsi-group/xpsi/blob/main/docs/source/intro.rst

drannawatts:
Removed the warning box (also updated our copyright years!!)

thjsal (approved):
I think it looks good now! I was just wondering why is the web page not properly updating when I run the build docs action manually. But I now realized it must be so because the action itself is checking out the main branch, and this update is not yet there. So I think we can just merge it and it should work after that.

PR #423, opened 2024-12-02
Update install.rst

denis-gc:


sguillot (changes requested):
Maybe say "and we are working on proposing a solution" instead of "and we are working to resolve them"

sguillot (approved):

Issue #320, opened 2023-08-16
Simplify the creation of a `CustomInstrument` from response matrice


sguillot:
The max/min input and max/min channels may be difficult to adjust as the choice of values can raise errors if some rows or columns of the matrix are filled with zero (or negative) values. By itself this is not a problem and this was implemented to save computing time.

The current error message tells the user to reduce the channels ranges to avoid this issue, without explaining the reason behind it. This could be clarified.

In practice, it shouldn't be necessary for the user to specify any range of max/min input and max/min channels.  The full matrix should be loaded, and then only the part of the matrix relevant to the input data should be use for the calculation.  

I think this could be transparent to the user (as it is currently for *all* X-ray spectral analysis packages), and therefore less complicated for the creation of the `CustomInstrument`, with perhaps some messages displayed to indicate the range of channels used.

Such improvement could be implemented at the same time as the FITS file implementation of the response matrices (#194)

Issue #138, opened 2022-09-22
Improve information on response matrix construction (CustomInstrument)


sguillot:
Help the user figure out how to solve the ResponsError:
```
Each row and column of the matrix must contain at least one positive number.
Try reducing the \"gap\" between the minimum input energy 
```

thjsal:
Is the suggestion to add this peace of text to the code (where?), or do you mean more information than this is needed?

sguillot:
Probably both in the code and the documentation.
Also, this should be solved with #194 and #320

lmauviard:
Closed by the PR #424

Issue #285, opened 2023-02-14
Make sure that all AssertionErrors in the Instrument class are explicit enough 


sguillot:
For example:
```
assert channel_array.shape[0] == self._matrix.shape[0]
```
gives
```
raise ChannelError('Channel numbers must be in a one-dimensional array of positive integers'
                   '(including zero).')
```

thjsal:
I see @lmauviard mentioned in the pull request linked above that this issue should be fixed now. Is that so?

However, I see that example mentioned above seems still the same as before:
https://github.com/xpsi-group/xpsi/blob/0fdbd56e07faa8bb19fb48e15d7ff1a7afeacc13/xpsi/Instrument.py#L301

lmauviard:
I fixed some others but not this one. I will fix them now.

PR #424, opened 2024-12-02
Add OGIP standard fits reading

### LINKED ISSUE #194: Provide FITS file implementation of the `CustomInstrument` classes
The CustomInstrument classes provided either in the tutorials or the Zenodo can be confusing. I think it could be greatly simplified by allowing the use to load directly from the FITS files. Response matrices (RMF and ARF) are standardised in X-ray astronomy, so there should not be any complications. It should be also relatively easy to extract the channels edges and to figure out from the file header if the RMF has already been multipled by the ARF or not.

This suggested feature should probably be tackled after the python3 release (since reading FITS file will rely on Astropy, and we will likely face difficulties with astropy version that are only compatible with Python 2)

lmauviard:
This update of the `Data` and `Instrument` class allows to load data from EVT and PHA file and response from ARF/RMF/RSP files.

For the `Instrument` : 
- Lines of 0 are removed automatically
- Removed the mandatory assertion of the channels to be less numerous than the energy edges
- It has been tested on NICER/XMM PN, MOS1, MOS2 ARF/RMF or RSP files

For the `Data` : 
- Allow to load background and compute the associated support
- Allow for phase shift of individual photons
- A generic plotting method is proposed to check the data

@yveskini Can we check the issue you had with this implementation, where you got a line full of random values in the Data ?

Closes #194


lmauviard:
Error arises from loading astropy during CI tests because it is not installed as a dependency there. I will fix this
I will also try to add it to an example notebook !

lmauviard:
Now also closes #320 because the channels to fit can be requested in the Signal.

lmauviard:
@sguillot This is done and needs reviewing

lmauviard:
Also closes #138

lmauviard:
And #285

sguillot (approved):

thjsal:
I  suspect that this pull request broke the likelihood checks in the TestRuns in the `examples/examples_modeling_tutorial`. They now fail since the data that they are fitting has been changed.

PR #436, opened 2024-12-02
Fixes to installation syntax warnings

thjsal:
I fixed the python3.12 syntax warnings when installing X-PSI by adding letter "r" before strings that contain the symbol "\". In addition, I removed the warning about  `xpsi.include` and `xpsi.include.rayXpanda` being absent from the packages by adding them to the package list.


DevarshiChoudhury (approved):
👍🏽

DevarshiChoudhury (approved):
Seems good

PR #438, opened 2024-12-03
Update contributions

sguillot:
and acknowledgements


drannawatts (changes requested):
Flip order of Rahul and Lami to preserve alphabetical order? Otherwise looks good!

sguillot:
good catch. I was too fast.

drannawatts (approved):
Oh yes good additional changes!

thjsal:
I updated also my contact information, and added Pushpita to the community.

PR #441, opened 2024-12-03
Change units of plots for RMF and ARF

sguillot:
Added color bar and units

corrected units in effective area plot and added labels


sguillot:
Other notebooks also need to be updated

denis-gc (approved):

PR #447, opened 2024-12-03
Add background plot

### LINKED ISSUE #91: Background plotter
Add documentation and tutorial on the background plotter

lmauviard:
Closes #91 

This is a tool to plot the spectrum of the background (for a given parameter if needed) and its distribution. This is plotted along the data and background support if it exists.

It comes with a few useful tools and require an imported model as an input, where it will fetch the instrument you are looking for from its name. Do you think this is the way to handle this ? Does everyone have namespaces like this for the signals ? I should be able to handle a few case but not all


thjsal (approved):
Looks good to me!

PR #449, opened 2024-12-03
400  docstrings mesh tools eff grav

MHoogkamer:
Added docstring for three cython functions related to the oblateness approximation implemented in X-PSI from Morsink et al. (2007)


thjsal (approved):

thjsal:
Looks good to me.

Issue #429, opened 2024-12-02
Update Stokes plot in tutorial for polarization


denis-gc:
Perhaps update the Stokes plots in the tutorial for polarization. Plot Q/I (and U/I), which varies between -1 and +1.

denis-gc:
Since the intensity drops to zero at some phases, the quantities Q/I and U/I become undefined. So better to keep the plots in their current form but correcting the normalization as I/max(I), Q/(max(|Q|), and U/max(|U|)

PR #455, opened 2024-12-04
#429 updated Stokes plots in polarization tutorial

denis-gc:
Updated normalization of the Stokes parameters,  y-axis labels and legends


thjsal (approved):
Looks good to me!

PR #456, opened 2024-12-04
Clusters plots with the residuals

pstammler:
I extended the residual Python code, so we can study the clustering effects from the residuals. The GridSpec has been changed, so you have to be cautious about how you want to display some coordinates in both axis. I also implement new parameters, like the threshold for the residuals, set by default to 2. Here is an example of the structure you may have to try in the Post-Processing tutorial notebook: """
pp = xpsi.SignalPlotter([ST.runs])
# Setting next the yticks for the 5 plots, alternating the main plot and the color bar y-ticks. # (None if using automatic tikcs)
xpsi.ResidualPlot.declare_settings(yticks=[[10,30,100],None, [10,30,100],None, [10,30,100],None, [10,30,100], [10,30,100], None, None, None]) pp.plot(IDs=OrderedDict([('ST', ['run']),
                        ]),
        combine=False, # use these controls if more than one run for a posterior
        combine_all=False,
        force_combine=False,
        only_combined=False,
        force_cache=True,
        nsamples=100,
        plots = {'ST': xpsi.ResidualPlot(threshlim=1.0)})

pp.plots["ST"].fig
"""


lmauviard:
I'll look into it, since I also want to add code to this file

lmauviard:
I move the clustering to an optional argument to be able to still make the old residual plot. Can someone test this ? It does work on my laptop when I remove the expected counts from `__caching_targets__`. Otherwise it does not work on my laptop, I get `Invalid cache target`. Can someone else check ?

sguillot:
Fix the residual distribution (bottom left plot)
Try to get the x-axis limit from the color bar limit if possible

Issue #431, opened 2024-12-02
Make a load NSX function in the base `Photosphere`


lmauviard:
Will simplify the writing of a new function. Should be useful because we only implement something with this format

ckazantsev:
While doing that i'd add also an `atmosphere setter` for `Everywhere`

lmauviard:
Solved by #466

PR #466, opened 2024-12-05
431 photosphere nsx

ckazantsev:
from issue #431 : Added loading of NSX atmosphere model from a file in the default atmosphere setter, and an atmosphere setter for `Everywhere`


lmauviard:
Good to go

lmauviard (approved):

Issue #444, opened 2024-12-03
Units of RMF and ARF in other notebooks


sguillot:
InstrumentSynergy was corrected but other notebooks should also be fixed

thjsal:
@sguillot You mentioned in the this issue was fixed in the Plotting library pull request. We can then close this?

thjsal:
I checked that the units at least in Modeling Tutorial look good, and since you said in the abovementioned pull request that this issue should have been fixed, I will close this.

PR #467, opened 2024-12-05
Plotting Library

sguillot:
New library collecting all plotting functions for the tutorial notebooks


sguillot:
This PR can be reviewed (and merged), but not all tutorial notebooks have been adapted. They are still correct and running, they do not make use of the new plotting routines.

sguillot:
This also solves #444

thjsal (approved):
This should be now good. I fixed a few things (see the commit messages above), but now the tutorials updated in this branch are working for me (and the example synthetic data matches to those created in the tutorial).

PR #473, opened 2024-12-05
New tutorial on importance sampling with X-PSI

yveskini:


thjsal:
I think we should probably add this tutorial preferably to the docs (so no not to the examples folder), so that it will be visible also from the web pages. I can do move this to there. I am also now checking if everything works and looks fine.

drannawatts (changes requested):
Reviewing just the tutorial notebook:

1. The line "For the new model, we changed the spin frequency from 314.0 Hz (which was used to generate the data) to 310.0 Hz (this is technically not the correct model, it's just an example)" - the final bit reads a bit strangely to me since if we changed frequency we'd also change the pulse profile.  Could we perhaps say instead something like:    

"[Note:  if we actually picked a different frequency in the model (perhaps because the true spin is uncertain), we would technically have to re-generate the pulse profile using that frequency as well.  Here we do not do that, which is equivalent to assuming that the data set is short enough that the change in frequency has not affected the pulse profile.  We've done this just because it makes a nice and fast example to show! "  

Feel free to rephrase as you like!

2. "new posteriors looks like" -> "new posteriors look like"

3. Not sure what you mean by "(look density method)"?  Also there needs to be a full stop at the end of that paragraph. 

Otherwise looks good to me!

thjsal:
I merged the main main branch to this branch, and the importance sampling notebook seems to be not working anymore. Need to fix some prefixes at least, since Lucien renamed them from "hot" to "p".

yveskini:
@drannawatts 
For comment 1, how about giving a reason for wanting to change only the frequency? Something like this:

"One reason to adjust only the frequency in the model while keeping the data unchanged could arise from inadvertently using a frequency different from the one used to fold the data, only to realize this discrepancy after performing the run (with the wrong frequency)"

Comment 2: Fixed.

drannawatts:
I think that suggestion for Comment 1 is fine!

thjsal:
I made a few sentence changes (e.g. clarifying what importance sampling means ... hopefully it is correct?). Before moving this to the Documentation pages, I think it is still good to increase the font sizes in the Corner plots and clarify what is inside `xpsi/utilities/ImportanceSample.py`.  I guess it is an alternative tool to use importance sampling developed by Serena. Maybe it is a bit weird that we are not showing how to use that in the tutorial? 

I can later try to fix these if you want.

thjsal:
I am working on this now.

drannawatts (commented):
I think this is looking good!  Is it ready for approval or are people still making last changes?

thjsal:
Not ready! I just checked how it would look like in the docs, and it is a mess!

thjsal:
Need to make changes to the notebook titles, subtitles etc. so that they don't show up randomly in the docs and in the index.

thjsal (approved):
I think this should be now more or less fine. The final change was to use * symbols instead of # to boldface normal text in the notebook. Otherwise, sphinx will treat those as titles and documentation becomes messy.

drannawatts (approved):

PR #476, opened 2024-12-05
Options to allow spherical star (but still not ignore time delays)

thjsal:
Here an option to allow spherical stars using an argument for the Spacetime class. Option for time delays could still be added. However, maybe could be skipped for now, since this would need to be added to all the various integrator files and  therefore introducing more and more repeated code. Maybe better to first (at some point) try to combine the integrators to avoid so much repetition.


MHoogkamer (approved):
Looks good to me! Clear implementation and works well

Issue #454, opened 2024-12-04
Implement Q-U plot in the polarization tutorial


denis-gc:
In the polarization tutorial, implement plot of the Stokes parameter in the Q-U plane, like Figure 4 of Słowikowska et al. 2009

https://ui.adsabs.harvard.edu/abs/2009MNRAS.397..103S/abstract

PR #479, opened 2024-12-06
#454 Q-U plot plus a few sentences on polarization notebook

denis-gc:


thjsal (approved):
I slightly modified a couple of sentences and fixed one link to the TestRun example (which was wrong already before). Looks good to me.

PR #482, opened 2024-12-06
254 unit tests for hotregion

Basdorsman:
Here I added basic run, getter and setter tests for all functions in hotregion.py

However, not every reasonable model configuration has been tested. For example, I am not testing ceding and omitting hot spots and numerical atmospheres yet. Best would be to also add some asserts to ensure that the correct values come out. Still, the tests we have so far can be checked and merged in my opinion.

Here I also deprecate the `is_secondary` keyword argument in the HotRegion class. So if this issue is closed that other issue can also be closed.

### DIFF
--- a/docs/source/Hot_region_complexity.ipynb
+++ b/docs/source/Hot_region_complexity.ipynb
@@ -194,7 +194,7 @@
     "                              num_leaves=100,\n",
     "                              num_rays=200,\n",
     "                              do_fast=False,\n",
-    "                              is_secondary=True,\n",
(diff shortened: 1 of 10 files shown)


Basdorsman:
Apparently in Python 3.8 and 3.9 it fails due to a | used as "or" in isinstance()...

DevarshiChoudhury (changes requested):
Add commit changing examples where `is_secondary` is used

DevarshiChoudhury (approved):
Looks good

PR #483, opened 2024-12-06
459 residuals plotting options

### LINKED ISSUE #90: Chi-squared calculator
Add a tutorial on the Chi-squared calculator (and the chi-squared calculator itself :-D )

### LINKED ISSUE #459: Add more plotting to residuals
Add the bolometric pulse (closes #90), blurring of the residuals and plotting of single parameter vector.

lmauviard:
More plotting options including : 

Choice of a parameter for the plots using postprocessor (instead of random samples)
Bolometric pulse : data and model along with chi square 
Blurring the residuals for better visualisation
Add sin_inclination to be used in ProjectionTool of required

PostProcessing notebook modified too

May close #90
Closes #459


lmauviard:
I will make an independent chi square calculator in Signal ! Wait until merging

lmauviard:
Now closes #90

lmauviard:
It is good to go now, can someone review this ?

sguillot (changes requested):
This looks all good, but one of the spectrum figures has a weird yrange:
Screenshot 2024-12-12 at 16 29 58
Is the yrange hard-coded somewhere ?

lmauviard:
I will have a look at it, this is a bit weird indeed

thjsal:
Seems like something the change to this yrange has occured already outside of this branch. In the online pages the figure looks still normal, when ran the PostProcessing notebook in another branch that is up-to-date with the main, I get this same.

sguillot:
Then perhaps we can ignore this very minor issue and close this PR.
The y-range for this plot can be fixed later no?

sguillot (approved):

drannawatts:
So before merging make a non-urgent issue to address the weird y-range at some point?

sguillot:
I created #495 to remember to make the fix later.
The branch can be closed I think

sguillot:
Sorry, I cancelled the review because the two residual plots show best-fit gaussians that are different. Even the distributions are different. Different binning? Not sure why. See figures:
Screenshot 2024-12-13 at 12 39 37
Screenshot 2024-12-13 at 12 39 27

lmauviard:
This is because the samples used for residuals are not the same. Basically you draw 100 different samples for each, which are different for each. If you use same `parameters_vector` you should get the same residuals

DevarshiChoudhury:
Closes #496

sguillot (approved):

xpsi/utilities/ProjectionTool.py
```diff
@@ -280,21 +281,46 @@ def check_model_param_names():
...
+            print ("ERROR: model not recognised. If model has derived quantities and does not fall into the categories mentioned below: transform your dictionary to match \"-U\" requirements. Recognised hot spot names are \"ST\", \"\PST\", \"\CST\", \"EST\", \"PDT\", \"CDT\", \"EDT\"; for two hot spot model connect each hot spot model with \"+\" (first for primary; second for secondary). If the two hot spots have the same name, add \"-S\" (if all the secondary hot spot is antipodal and its temperature and radius is the same of the primary); \"-U\" if the properties of the secondary spot is completely unrelated to the primary ones; \"-Ua\" if the secondary hot spot is antipodal to the primary but has independent radius and temperature.")
```
thjsal, line 290: Why did you remove the letter "r" from here? There should be "r" so that it won't give a SyntaxWarning, since having a string that has backslashes?

thjsal: I added the "r" now back.

xpsi/PostProcessing/_residual.py
```diff
@@ -293,7 +372,7 @@ def _add_data(self):
 
         data = self._ax_data.pcolormesh(self._signal.data.phases,
                                         channel_edges,
-                                        self._signal.data.counts/2.0,
```
thjsal, line 296: Since not dividing with 2 anymore, should the title in the plots change from `counts` to `counts/cycles`? This change means also that the old X-PSI signal plots will not look the same anymore if produced using the same script. Maybe that is at least worth to mention in the tutorial or in the Changelog?

drannawatts: Agreed, 'counts/cycle' (well actually it's counts per bin per cycle I guess)?  and agreed we should mention this explicitly in the Changelog

thjsal: Yeah, I guess it is maybe counts / (bin / cycles)... if you count all the bins within all the shown cycles. But I am not sure what is best/correct. Would be just good to be clear that this has changed.

thjsal (approved):

xpsi/utilities/ProjectionTool.py
lmauviard: Sorry I did not remove it, I guess it was already not in the original branch when I pulled to create this one

PR #492, opened 2024-12-06
Update years and stop assigning all tasks to Tom by default

drannawatts:
Update copyright years and remove Tom as default assignee for features!


thjsal (approved):
Looks good!

PR #493, opened 2024-12-06
Example scripts fixed

thjsal:
Example scripts in `examples_modeling_tutorial` folder were updated to pass the likelihood checks using the new synthetic data. In case of `TestRun_BB.py` I updated the model and instrument response methods to match those of the actual Modeling tutorial notebook in the documentation pages. In the other examples, I just updated the likelihood values and added a comment that the model (and the response files) used in those examples are not the same as used when generating the data (which was the case already before anyway, since they are using a numerical atmosphere).


lmauviard (approved):
Good for me

PR #494, opened 2024-12-09
Corrected deepcopy bug on instrument and data

### LINKED ISSUE #491: Derive class not working for MOS2
alpha for MOS2 return None instead of being bound to  MOS1

class derive(Derive):
    global MOS1

    def __init__(self):
        pass

    def __call__(self, boundto, caller=None):
        return MOS1.instrument['alpha']

lmauviard:
Closes #491 

Corrected the bug originating from the deepcopy in the `Signal.py` 
I have no run on hand with this derive setup, can anyone check this actually fixes the bug ?


thjsal (approved):
I checked that this fixes the bug for me at least.

PR #498, opened 2024-12-16
Removed name from instrument and its dependencies

lmauviard:
I removed the `name` parameter from the `Instrument` init which cause the merging of the scaling factor into the instrument to break if no `name` parameter was given.


thjsal (approved):
Looks good to me. I also double checked that this solves the same error for me too, and the notebooks plotting arfs are working.

PR #499, opened 2024-12-16
Changelog and version number updates

thjsal:
More text still to be added to the Changelog.


MHoogkamer (changes requested):
"Updated the installation instructions for Helios and the example job script." should be "Helios and Snellius". Otherwise looks good to me!

thjsal:
> "Updated the installation instructions for Helios and the example job script." should be "Helios and Snellius". Otherwise looks good to me!

Sure. I have now added that.

MHoogkamer (approved):

thjsal:
I have now rerun 2 tutorials to the update version number in them. Also fixed the SBI notebook to use the proper path to the example main module, and removed the sentence `Make sure to run this first (python examples/examples_fast/Modules.main.py), to get all output files needed to execute this notebook.`, because that is not needed (all the files are already in GitHub).

drannawatts (approved):

PR #505, opened 2025-01-03
Import order issue moved to FAQ.

thjsal:
As discussed during the hack week, I added information about the import order issue of numpy and pymultinest to the FAQ pages so that the GitHub issue can be closed. 

I also tried to reproduce the issue with newer X-PSI installations (both X-PSI v3.0.0 and v1.2.1 with newest dependencies), but I was not able to do that. Instead, the import order did not seem to affect the results at all (which I think makes more sense). I mentioned this also in the FAQ. I could not either reproduce the issue with the exact packages that were reported in the original issue, since my current laptop does not have an old enough version of gcc.


drannawatts (approved):
Looks good to me!

PR #506, opened 2025-01-08
420 accretion disk tutorial

Basdorsman:
The tutorial is added and ready for checking. I compiled the documentation locally and that appeared to work fine.


Basdorsman:
@thjsal any idea why this test is failing?

thjsal:
I have do not know why this happens (at least right now), but this seems to be the error message that comes when GitHub is installing X-PSI:

```
  File "/home/runner/work/xpsi/xpsi/setup.py", line 77, in <module>
Operating system: linux
GSL version: 2.7.1
Warning: the rayXpanda package cannot be imported. Using fallback implementation.
    import rayXpanda
ModuleNotFoundError: No module named 'rayXpanda'

During handling of the above exception, another exception occurred:

  File "/home/runner/work/xpsi/xpsi/setup.py", line 82, in <module>
    sub.call(['%s'%CC,
  File "/home/runner/micromamba/envs/xpsipy3/lib/python3.[9](https://github.com/xpsi-group/xpsi/actions/runs/12669137815/job/35306004548#step:8:10)/subprocess.py", line 349, in call
    with Popen(*popenargs, **kwargs) as p:
  File "/home/runner/micromamba/envs/xpsipy3/lib/python3.9/subprocess.py", line 951, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/home/runner/micromamba/envs/xpsipy3/lib/python3.9/subprocess.py", line 1837, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'gcc-[10](https://github.com/xpsi-group/xpsi/actions/runs/12669137815/job/35306004548#step:8:11)'
```

thjsal:
Probably some default packages/compilers that GitHub uses have been upgraded, and the new versions fail for some reason.

drannawatts:
Gcc issue associated with the new ubuntu version? Seems like the one they are in the middle of flipping to as ubuntu-latest uses gcc-11 or 12 rather than 10 (since it's not finding gcc-10)

thjsal:
I guess so. Then probably just need to change this line (and possibly also in other workflows):
https://github.com/xpsi-group/xpsi/blob/28013876962c1e1ad74346064db10cf4e0414619/.github/workflows/ci_tests.yml#L57

drannawatts:
Looks like we'll have to go to gcc-12...

https://github.com/actions/runner-images/issues/10636

drannawatts (commented):
I've corrected a few small typos. Looks good though! 

I think Tuomo's suggestions are good.  I also wondered whether there is any way to have hyperlinks to the various papers that are cited, in the notebooks? Maybe this is not easy though. 

You mention in the middle of the tutorial the 'distance to the disk'. We are assuming that this is identical in our modelling to distance to star, I think? I'd suggest adding a note to make this clear.

Basdorsman:
Just to document my considerations: I was considering changing this notebook to include the accretion disk as a separate item in the list of `photospere.signals`, i.e. the same way how now two hot spots occupy separate items in the signals list. The benefit of that is to easily access the disk spectrum counts later on (currently you'd need to compute star without disk and then subtract it by hand). For now I've decided to leave it as is, because when it is separated, sampling and post-processing breaks. For future work, I'd recommend to implement the disk outside the `star` and have it be a separate class that can contribute to the signal alongside the `star`.

thjsal (approved):

Issue #468, opened 2024-12-05
Verify and implement the replacement of less_than_llzero with random_near_llzero in likelihood.py


Basdorsman:
We have found that the only line in likelihood.py where less_than_zero is used, can be replaced by using random_near_zero instead. This addressed a problem we experienced, where the sampler got stuck in an infinite loop. This is a single line change but we'd like to make a serious attempt to verify first:

- Why did this problem appear only in recent runs?
- Why was this implemented in this way to begin with?
- Does removing this affect the efficiency of the sampling algorithm or otherwise affect intended behaviour?
- Are posteriors unaffected? Should this change be noted for old or new published results?

We aim to address these questions by consulting the literature of the multinest and x-psi algorithms, their implementations, and doing some comparison runs.

Basdorsman:
Why did this problem appear only in recent runs?

- I am unsure, it seems like this problem was always lurking, waiting to happen. 

Why did this happen to begin with? Are ellipsoids drawn fully outside multidimensional prior boundary? Why?

- MultiNest is not aware of multidimensional prior boundaries, so there is no reason why it wouldn’t draw those ellipsoids. A more complete answer would go in depth about the algorithm of how ellipsoids are drawn. 

How is the multinest algorithm affected by this workaround? Any unintended behaviour? Any algorithmic efficiency loss?

- Instead of outright rejection, points will now be weakly rejected when their loglikelihoods are still lower than the likelihoods of the current livepoints.
- The bulk of the work is in the evaluation of the loglikelihoods, so rejection after loglikelihood comparison amounts to negligible extra work.

Are previous and new scientific results safeguarded? Is the old point “rejection” method and current fix allowed under Bayesian statistics? E.g. between new and old, is sampling biased or are posteriors affected?

- The exploration of parameter space is only different at the stage where some livepoints are random_near_llzero, the posterior probability distributions should not be significantly affected by this. Further along, points will be rejected. 

Are posteriors unaffected? Should this change be noted for old or new published results?

- We find that posteriors are unaffected in multiple cross-tests (my own accretion disk model and Tuomo's polarization runs).

Further discussion: Is it feasible to absorb all prior multidimensional boundaries in inverse sample, and completely clean prior.call.
- While this is cleaner and makes MultiNest explicitly aware of the boundaries, it is not straightforward to implement this especially for complex and multidimensional boundaries. The efficiency gain is limited, since the time spent on the out-of-bounds points is very limited anyways.

PR #507, opened 2025-01-08
implemented random_near_llzero

Basdorsman:
See also associated issue [https://github.com/xpsi-group/xpsi/issues/468](url).


thjsal:
Your link to the associated issue is not working. It should be this: https://github.com/xpsi-group/xpsi/issues/468

I am fine with this change, but I suggest to also change X-PSI version number to 3.0.1 and mention this change in the Changelog.

Basdorsman:
By the way, I think we can remove likelihood.less_than_llzero() completely, is that right?

thjsal (approved):

thjsal:
I changed the version number also in `docs/source/conf.py` and corrected a small typo in the Changelog. I think this looks good. And indeed, maybe `likelihood.less_than_llzero()` could be removed completely.  Unless you think it is worth to keep it there if somebody still would use the function in their customised codes or tests (but sure if having a `CustomLikelihood.py` you can add this function there yourself as well).

PR #509, opened 2025-01-09
Updating GitHub workflows to use newer gcc

thjsal:
As discussed in https://github.com/xpsi-group/xpsi/pull/506, newer gcc version is needed for the GitHub Action workflows when installing X-PSI, because the latest default Ubuntu version has upgraded and has no gcc-10 anymore.


thjsal:
Everything seems to be working now. I also added python version 3.12 to be one of your testing versions.

drannawatts (approved):
OK great!  Into the future we go.....

PR #513, opened 2025-01-16
Fixed Everywhere issues

### LINKED ISSUE #412: Possible bug in Likelihood of Photosphere Everywhere with time_invariant=True
When running the likelihood estimator (likelihood.check) with `Everywhere` and time_invariant=True, different executions give different values of likelihood, either close to llzero or close to the true value of -1.8e3. Doing the exact same thing with time_invariant=False always gives the correct value.

### LINKED ISSUE #512: Numerical atmosphere working incorrectly with Everywhere
It appears that the numerical atmosphere option using the current standard Everywhere class produces largely incorrect signals. The reason is the introdcution of the ``dummy phase_shift`` parameter in v2.2.4 of X-PSI that was done to enable likelihood calculations with Everywhere objects. Thus, now the default `_compute_cellParamVecs()` function in the Everywhere class expects that the atmosphere data have also phase shift in it (it is one of the parameters in `self.vector`):
https://github.com/xpsi-group/xpsi/blob/368f5f8e86f81bd895f0d4920ab53a0833863eec/xpsi/Everywhere.py#L452

Nothing breaks if no phase_shift column for atmosphere data are provided, but the computed signal appear vastly wrong. One workaround is to modify the `_compute_cellParamVecs()` function to the following (that only allows to use the first Everywhere parameter, i.e. temperature, and effective gravity as atmosphere parameters):

```
  def _compute_cellParamVecs(self):
      """
      Precompute photospheric source radiation field parameter vectors
      cell-by-cell. Free model parameters and derived (fixed) variables can
      be transformed into local comoving radiation field variables.

      Subclass and overwrite with custom functionality if you desire.

      """
      # all radiate, but can be changed with overwrite
      self._cellRadiates = _np.ones(self._theta.shape, dtype=_np.int32)

      self._cellParamVecs = _np.ones((self._theta.shape[0],
                                      self._theta.shape[1],
                                      2),
                                      #len(self.vector)+1),
                                     dtype=_np.double)

      #self._cellParamVecs[...,:-1] *= _np.array(self.vector)
      self._cellParamVecs[...,:-1] *= _np.array([self.vector[0]])

      for i in range(self._cellParamVecs.shape[1]):
          self._cellParamVecs[:,i,-1] *= self._effGrav
```

lmauviard:
Solves both problems : 
Closes #412 : likelihood computation with time_invariant=True
Closes #512 : Numerical atmosphere implementation

Integration was slightly changed to match the same format as for elsewhere.


thjsal:
I suggest to update this branch with main and then change the version number to 3.0.2 and summarise in Changelog what has been changed.

thjsal:
I was hoping to do that myself, but modifying a pull request from a fork seems still too difficult for me.

lmauviard:
Ok Tuomo, I will do it then !

lmauviard:
@thjsal Just did it !

thjsal:
Looks good to me! I just added one comment in the docstring of cellParamVecs in Everywhere and removed some unnecessary titles from the Changelog and fixed one typo (now it was easy to modify since I could use the browser for that as it was one of the modified files :D).

PR #515, opened 2025-01-21
Added posterior predictive error on pulse model plotting

lmauviard:
I added posterior predictive error bars to the modeled pulse profile plot (`_residuals.py`). 

As before, I compute `nsamples` models and display the mean as the model. But now, I do not put the error bars on the data but I rather compute `n_realisation_per_model` (default=1) Poisson realizations of every model and use this to compute the error bars on the modeled pulse. 
In the case with little realizations (`n_realisation_per_model` x `nsmaples` < 100), I plot the Poisson error on the modeled pulse instead. This is especially useful when the user want to do the plot with a given parameter vector.

Chi square is now no longer plotted on the pulse profile as it is not very relevant. But it can be accessed by `plot.bolometric_chi2` if the residual plotter was run once.

The PostProcessing notebook was updated accordingly.


thjsal:
I get
`AttributeError: 'dict' object has no attribute 'bolometric_chi2`
if trying to access the bolometric chi square in the way you said. Maybe the way to access it could be shown in the tutorial?

Quickly looking looks otherwise fine to me. Except that the legend text for "Data" and "Model" is on top of the curves. Can it be easily moved (e.g. for corner plot we can usually give `legend_corner_coords`)?

thjsal:
I also found it a bit confusing that you call different samples as different models. Would `n_realisation_per_sample` be a better name than `n_realisation_per_model`?

lmauviard:
I did the changes you mentioned with bolometric chi2, n_realisation_per_sample and label positionning. I added a way to plot labels for non diagonal regions of posteriors if needed.

The PostProcessing notebook has been updated

lmauviard:
Solves #522

thjsal (approved):

PR #517, opened 2025-01-28
Fixes #516

### LINKED ISSUE #516: `_poisson_likelihood_given_background.pyx` likelihood computation failure for phase bins with 0 counts
https://github.com/xpsi-group/xpsi/blob/2844e52c3c915ff34a25fc0a8daf5d1997701ec2/xpsi/likelihoods/_poisson_likelihood_given_background.pyx#L202

#L201 and #L202 in the above file ends up with log(0) if the model outputs a phase with 0 counts thereby crashing the likelihood computation

DevarshiChoudhury:
Following scenarios are addressed by the code changes:
- `data counts = 0` but model `EXPEC>0`. The `LOGLIKE` won't crash anymore and would have a negative value since it enters the first `if` block. 
- `counts = 0` and `EXPEC = 0`: Then `LOGLIKE=0`.
- `counts > 0` and `EXPEC = 0`: Then `LOGLIKE` is equivalent to a `random_near_llzero`, thus penalizing this situation.
- `counts > 0` and `EXPEC > 0`: Then the `if` block is executed with both lines contributing to the `LOGLIKE` value.

### DIFF
--- a/xpsi/likelihoods/_poisson_likelihood_given_background.pyx
+++ b/xpsi/likelihoods/_poisson_likelihood_given_background.pyx
@@ -198,8 +198,20 @@ def poisson_likelihood_given_background(double exposure_time,
     for i in range(<size_t>STAR.shape[0]):
         for j in range(<size_t>STAR.shape[1]):
             EXPEC = (STAR[i,j] + background[i,j]/n) * exposure_time
-            LOGLIKE -= EXPEC
-            LOGLIKE += counts[i,j] * log(EXPEC)
+
+            # Ensuring that the log likelihood doesn't crash
+            # even when model counts is zero in a given phase bin.
+            if(EXPEC > 0.0):
+                LOGLIKE -= EXPEC
+                LOGLIKE += counts[i,j] * log(EXPEC)
+    
+            elif(counts[i,j] == 0 and EXPEC == 0):
+                pass
+            
+            # Penalizing situations where the data counts are not zero 
+            # but the model counts are.
+            else:
+                LOGLIKE = -1.0e90 * (0.1 + 0.9*np.random.rand(1))
 
             STAR[i,j] += background[i,j]/n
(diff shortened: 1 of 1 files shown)


xpsi/likelihoods/_poisson_likelihood_given_background.pyx
```diff
@@ -198,8 +198,20 @@ def poisson_likelihood_given_background(double exposure_time,
...
+                LOGLIKE -= EXPEC
+                LOGLIKE += counts[i,j] * log(EXPEC)
+    
+            elif(counts[i,j] == 0 and EXPEC == 0):
+                pass
+            
+            # Penalizing situations where the data counts are not zero 
+            # but the model counts are.
+            else:
+                LOGLIKE = -1.0e90 * (0.1 + 0.9*np.random.rand(1))
```
thjsal, line 214: Just a small comment, but would it perhaps make more sense to have 
`LOGLIKE += -1.0e90 * (0.1 + 0.9*np.random.rand(1))`?

Currently, it will add log-likelihoods only from those bins that happen to come after this condition, but from those that are before.

Or break out from the loop when having this condition so that likelihoods from other phase-energy bins won't get added on top of this (although then the star and background signals are not saved for all bins)?

DevarshiChoudhury: Yea, I think adding it to `LOGLIKE` should be okay. I'll add that change.

thjsal:
Looks good to me. Should we also change the X-PSI version to 3.0.3 to add a comment about this change in Changelog?

DevarshiChoudhury:
I was thinking about that indeed. Let me do that.

thjsal:
I made some small changes to the Changelog text. The version number should be updated also in `setup.py` and in `docs/source/conf.py`.

thjsal (approved):

PR #519, opened 2025-01-31
Everywhere with time_invariant=True, phases and leaves forced to 1

ckazantsev:
phases and leaves forced to 1 in Everywhere if `time_invariant=True` and interpolation initialization skipped if phases/leaves are 1


thjsal:
I am changing the branch to be merged from main to a new "everywhere_problems" branch that I just created, so that it is much easier for me to checkout and commit any possible changes to it. I hope that is ok? We can then create a new pull request from that branch to main ....

Issue #520, opened 2025-02-06
[Feature] Using `LatinHypercube` for `Prior.draw`


DevarshiChoudhury:
**Is your feature request related to a problem? Please describe.**
Currently we use `np.random.rand` to perform draws from prior which may exhibit clustering, thereby potentially biasing SBI training if the `Prior.draw` function is used for training data generation.

**Describe the solution you'd like**
Using Latin Hypercube Sampling ensures that the samples are spread more evenly across the entire domain, making it a better choice for high-dimensional spaces when compared to simple random sampling.

sguillot:
Should we had this to Hack Week Spring Clean ?

The PR just needs to be reviewed and approved, right ?

thjsal:
I have reviewed it but I understood that you decided in one of the meetings that you want to use the newer Latin Hypercube syntax that is not compatible with older Python3 versions. The current pull request is now using the older syntax that still works for all Python3 versions.  If using that is fine, I think it is ready to be merged (I think that would be ok). If using the newer syntax, we should update also the GitHub actions and X-PSI installation instructions to use/mention newer than the new minumum Python version.

PR #521, opened 2025-02-06
Using Latin hypercube sampler for `Prior.draw`

DevarshiChoudhury:
Addresses feature request #520.

I have checked its expected application on the modelling notebook where we print the outputs of `Prior.draw`. The prior distributions looked ~identical, perhaps even slightly more uniformly distributed than before as seen by eye.


thjsal (commented):
Looks good to me. However, I find one thing worse with change. Looks like you cannot re-produce exactly the example results anymore every time you run them. I guess that is because you should fix somehow the seed of this new Latin hypercube sampler.

For example, if running TestRun_BB.py twice in a row, I get first time:

```
The support occupies an estimated 25.4% of the hypervolume within the unit hypercube...
Fractional hypervolume estimated.
Sampling efficiency set to: 3.1504.
```

And second time:

```
The support occupies an estimated 24.8% of the hypervolume within the unit hypercube...
Fractional hypervolume estimated.
Sampling efficiency set to: 3.2264.
```

Previously, fixing the numpy seed in the main script was enough to guarantee the numbers won't change here when executing several times.

thjsal:
Perhaps you could fix the seed for this hypercube sampler in all the example files, so that they remain very re-producible. Adding something similar to this line:
https://github.com/xpsi-group/xpsi/blob/d38db0bd07d048b2d47c053cedbcc94484f1e11e/examples/examples_modeling_tutorial/TestRun_BB.py#L16

DevarshiChoudhury:
@thjsal it wasn't as straightforward a fix. I've added a `LHS_seed` arg in `Prior.draw`, `Prior.estimate_hypercube_frac` and `Prior.unit_hypercube_frac`. A `LHS_seed` is then added in `runtime_params` dict in the example files for `Sample.nested` to parse as a kwarg, passing it to `NestedSampler` as a kwarg which then uses it to call `Prior.unit_hypercube_frac`. I think this should make the examples fully reproducible although I haven't tested it myself. Also, if no `LHS_seed` is passed, it'll just default to a random seed.

thjsal:
Seems like it is not working for me. I get first (when running TestRun_BB.py):
```
  File "/home/saltuomo/.conda/envs/xpsi_py3/lib/python3.12/site-packages/xpsi-3.0.5-py3.12-linux-x86_64.egg/xpsi/NestedSampler.py", line 58, in __call__
    print("LSH_seed: ",LHS_seed)
                       ^^^^^^^^
UnboundLocalError: cannot access local variable 'LHS_seed' where it is not associated with a value
```

That I can fix by doing `LHS_seed = kwargs.setdefault('LHS_seed', 0)` instead of `LHS_seed = LHS_seed if LHS_seed is not None else None`

Then I get 
```
  File "/home/saltuomo/.conda/envs/xpsi_py3/lib/python3.12/site-packages/xpsi-3.0.5-py3.12-linux-x86_64.egg/xpsi/Prior.py", line 235, in draw
    sampler = qmc.LatinHypercube(d=len(self), 
              ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: LatinHypercube.__init__() got an unexpected keyword argument 'rng'
```

Seems like the "rng" input parameter was used to be named "seed" in Scipy version before 1.15. Do you think we should force X-PSI to use Scipy versions newer than this (it also requires Python newer than 3.10)? Anyway, I changed that to "seed", but now I get:

```
  File "/home/saltuomo/.conda/envs/xpsi_py3/lib/python3.12/site-packages/xpsi-3.0.5-py3.12-linux-x86_64.egg/xpsi/NestedSampler.py", line 62, in __call__
    kwargs['sampling_efficiency'] /= self._prior.unit_hypercube_frac(LHS_seed=LHS_seed)
                                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
TypeError: 'float' object is not callable
```

I don't know yet how to solve this one.

DevarshiChoudhury:
Thanks for fixing the `dict` related issue. I had assumed that since the env files only mention scipy and not a version number, it would automatically install the latest version (v1.15.2) but I suppose not since I have v1.12.0 in my system. Do you envision any problem associated any other problem with updating to v1.15 for Scipy?

However, looking at your error, it also seems that the error is taking place at an earlier/later point, i.e. at `self._prior.unit_hypercube_frac(LHS_seed=LHS_seed)` and not `qmc.LatinHypercube(d=len(self)`, so different from where the `rng` was getting stuck. 

It's not immediately obvious to me why this is either. We can come back to this if it's not trivial

thjsal:
I think it may use the latest Scipy version if creating the X-PSI environment now. But probably the version v1.15 was not yet available when we last time created that environment. I don't know if upgrading to that version would cause issues. Looks like at least the Github CI-tests are already using that version without issues. But would be good to add another CI-test for drawing samples, and calling the sampler (this is not tested currently). If we do that, we should also stop testing Python 3.8 and Python 3.9 as those would not work anymore. Alternatively, we could call the parameter "seed", since it should still work, probably just giving a deprecation warning for now.

Regarding the latest error, I found that is caused because giving arguments for a property function is not allowed. The error goes away if not declaring the `unit_hypercube_frac()` as a property function in `Prior.py`. 

Next error is from pymultinest which is not happy that it got an additional `LHS_seed` parameter.

  File "/home/saltuomo/.conda/envs/xpsi_py3/lib/python3.12/site-packages/xpsi-3.0.5-py3.12-linux-x86_64.egg/xpsi/NestedSampler.py", line 72, in __call__
    _ = pymultinest.solve(self._likelihood,
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/saltuomo/.conda/envs/xpsi_py3/lib/python3.12/site-packages/pymultinest/solve.py", line 71, in solve
    run(**kwargs)
TypeError: run() got an unexpected keyword argument 'LHS_seed'

thjsal:
Okay, pushed now fixes to all the errors mentioned above. The last one (pymultinest error) was solved by just not having LHS_seed defined in the kwargs that are sent to pymultinest.

Now the test runs seem to work as intended. But perhaps should still check/think if the removal of the "@property" definition `unit_hypercube_frac()` function was completely fine, and whether we should force to upgrade to a newer Scipy.

I am also planning to set up (maybe in a couple of days) a couple of new Github CI-tests to make sure the sampling works also with newest packages.

thjsal:
I also changed `LHS_seed=LHS_seed if LHS_seed is not None else None` to simply `LHS_seed=LHS_seed`, since I could not understand why the longer version would be needed (it just sets the seed to None if it already is None).

DevarshiChoudhury:
So if I understand it correctly, right now the argument used is `seed` and that works across versions at the moment. So let's retain that for the time being. Once it's fully deprecated, we can update to `rng`. I think we can perhaps merge this pull request now then @thjsal?

thjsal (approved):

thjsal:
Sounds good. I suggest to "squash and merge".

PR #525, opened 2025-02-20
Continuous integration tests for sampling

thjsal:
I added a couple of new GitHub CI tests checking that

1) `unit_hypercube_frac` returns the same value as before.

2) Sampling with UltraNest does not crash. To make this test work I needed to add a requirement that GitHub installs UltraNest version that is newer than 4.0. Otherwise, test failed if using python3.9 or python3.10 (where it installed older UltraNest by default, which seemed inconsistent with the installed Numpy version).

I tried also adding a test for MultiNest, but could not in the end get the test program to find the MultiNest library files. However, I left the MultiNest installation and the corresponding test there as commented out if someone will figure out how to get it working.

### DIFF
--- a/.github/workflows/ci_tests.yml
+++ b/.github/workflows/ci_tests.yml
@@ -51,6 +51,7 @@ jobs:
         flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
     - name: multinest and pymultinest install
       run: |
+        sudo apt install mpich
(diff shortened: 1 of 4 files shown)


thjsal:
@DevarshiChoudhury 

Is it ok if I merge this to your Latin Hypercube branch? It possible to do without your approval, but just wondering if it is for you? 

This request just adds a new continous integration test that can be useful for that pull request.

Issue #527, opened 2025-03-10
Synthesise needs `num_phases` >4


ckazantsev:
Similar to a previous issue, when trying to synthesise data with `Everywhere` and `phase_invariant=True` (or False if  `num_phases` <5) it crashes. Probably comes from some interpolation (similar to PR #519 maybe?) but needs to be investigated

PR #528, opened 2025-03-14
phases and leaves forced to 1 (#519)

thjsal:
This is the pull request #519  moved to a branch within x-psi group for easier testing and modifications if needed.


thjsal:
I changed the version numbers and started to write something to the Changelog.

I also tested that this fix works for me too. But I also get that it crashes if creating synthetic data as you reported here #527 . I get this error:
```
gsl: interp.c:37: ERROR: insufficient number of points for interpolation type
Default GSL error handler invoked.
Aborted (core dumped)
```
Is it same you got?

I will a bit later try to find out how to also fix it.

thjsal:
Okay, I think it should be now fixed. I made the same changes to` synthesise.pyx` that had been made to `default_background_marginalisation.py` to skip phase interpolations if there is only one phase. It works for me now. I can later test more and update the changelog.

thjsal:
I tested different examples and tutorials and found that the Global surface emission -tutorial is not working anymore with changes done in this pull request (it crashes). Looks like there in `def plot_2D_pulse()`, it is expecting the number of phases to still be larger than 1 even when `everywhere.time_invariant = True`.

sguillot:
I guess an easy fix here is to modify plot_2d_pulse() to force a call of plot_1d_pulse() if it finds only 1 phase in the data/signal

thjsal:
Looks like the idea in the tutorial is though to get a figure like this (at least currently):

image

I can actually get it if replacing 
```
temp = np.tile(photosphere.signal[0][0], (1,len(everywhere.phases_in_cycles)))
plot_2D_pulse(temp,
              x=everywhere.phases_in_cycles,
              shift=np.array([0.0]),
              y=energies,
              num_rotations=2.0,
              ylabel=r'Energy (keV)')
```
with 
```
temp = np.tile(photosphere.signal[0][0], (1,100))
plot_2D_pulse(temp,
              x=np.linspace(0.0, 1.0, 100), 
              shift=np.array([0.0]),
              y=energies,
              num_rotations=2.0,
              ylabel=r'Energy (keV)')
```

So unlike before, we cannot use "fake phases" in `everywhere.phases_in_cycles` for plotting, but you need to create the phase dimension yourself if want make such a plot.

The next problem is that if you decide to switch to `everywhere.time_invariant = False` on fly (as in the tutorial now), you still have only 1 phase bin by default. So after switching you also need to also remember to upgrade the phase binning using e.g. `everywhere.set_phases(512,100,None)` (first number is num_leaves, and second num_phases).

With these changes the tutorial is running further.

ckazantsev:
The synthesize corrections do work for me, i had the same crash as you did and now it looks like everything works correctly. Thanks for taking care of that. As for the plotting comment, it does not really make sense to have these `plot_2d_pulse` if `time_invariant=True`, no? If we do want this plot to be shown, i would say we can use your corrections to fix the tutorial or modify the function to accept only 1-bin phase arrays (plot only one big bin in phases, if that's possible)

ckazantsev (approved):

thjsal:
Thanks for checking!. I'll just quickly update the Changelog and then merge.

thjsal:
Forgot to answer to this question:
`As for the plotting comment, it does not really make sense to have these plot_2d_pulse if time_invariant=True, no`

I agree it seems a bit unnecessary, but I think the point in the tutorial is probably just to show that indeed the signal is constant in phase. I don't mind dropping it out at some point, but would then need to think more how to rewrite that part of the tutorial.

PR #538, opened 2025-04-30
Pileup module addition

ckazantsev:


sguillot:
Instead of creating a new module file `Instrument_pileup.py`, wouldn't it be preferable to add a new class in the `Instrument.py` module ?
Unless this is not good practice

ckazantsev:
I'll be picky a bit but i have some corrections to add : 
1) "convolved to `spectral_model * ARF` and `RMF`." --> mathematically its `ARF * spectral_model` that is convolved, and `RMF` is applied after convolution and sum.
2) grade migration is the main fitted parameter, but psf_fraction CAN also be fitted. the rest are indeed always fixed (and have to be)
3) the last sentence make it sound like all other parameters are read from the header, which is not true. only fracexpo and exptime (frame time) are, the rest have to be given as values as usual in xpsi (but default values have been set if they are not given)

sguillot:
And now?
Screenshot 2025-05-14 at 15 12 46

sguillot (approved):

PR #548, opened 2025-05-12
547 Module generator: update generate.ini to load fits or txt files

denis-gc:
Now generate.ini contain a new argument "-format" which can be set to either "txt" or "fits"

Screenshot 2025-05-12 at 15 10 34


pstammler:
Other option:
parser.add_argument('--use-fits-format',
                    action='store_true',
                    help='Does the observational data are stored in fits-format files?',
                    comment_line_above='FITS observational data files?',
                    empty_lines_below=2,
                    comment=True)

Issue #544, opened 2025-05-12
SignalPlotter crashes in PostProc. notebook


sguillot:
Crash happens for the SignalPlotter:
```
pp = xpsi.SignalPlotter([ST.runs])
pp.plot()
```

With the following guilty line (from fgivenx)
```
File ~/Softwares/anaconda3/envs/xpsi/lib/python3.11/site-packages/fgivenx/plot.py:101, in plot(x, y, z, ax, **kwargs)
     99 # Rasterize contours (the rest of the figure stays in vector format)
    100 if rasterize_contours:
--> 101     for c in cbar.collections:
    102         c.set_rasterized(True)
    104 # Remove those annoying white lines

AttributeError: 'QuadContourSet' object has no attribute 'collections'
```

Could be matplotlib issue due to fgivenx using an older version matplotlib
https://github.com/matplotlib/matplotlib/issues/29575

sguillot:
This is an issue of compatibility of fgivenx with matplotlib.
We need to require `matplotlib==3.9.2`

sguillot:
Will be fixed with #524

sguillot:
Solved with https://github.com/xpsi-group/xpsi/commit/2881ce8e7a6f5bb28bd3d7ec5ce95a5f573cda22

PR #550, opened 2025-05-12
New environment.yml file

sguillot:
We now have a single environment.yml file, and the corresponding and updated documentation.

Also correct #544 (matplotlib==3.9.2 requirement).


thjsal:
Okay, looks like this new environment file is so specific that it only works with python3.9 version :D (see the GitHub tests above). I think we could try to allow more Python versions.

thjsal (approved):
I tested that this environment works in my laptop too.

However, I did not test these new requirements in a cluster. Quite many of the new required exact packages are probably not needed there if not doing post-processing.  In addition, in some clusters we used to use the native gsl instead of conda installed gsl. Hopefully, this will not make a difference. I will anyway approve this now, and perhaps it can be later fixed if needed.

EDIT: I later realized we are not even using the environment file in the HPC installation instructions, so there should be no issue due to the changes made here.

PR #554, opened 2025-05-12
537 save output as npy

MHoogkamer:
Now both MultiNest and UltraNest automatically produces .npy files for the .txt, dead-birth.txt, and phys_live-birth.txt files. Also shortened the line length for better readability.

### DIFF
--- a/xpsi/NestedSampler.py
+++ b/xpsi/NestedSampler.py
@@ -68,7 +68,7 @@ class NestedSampler(object):
                               log_zero = self._likelihood.llzero,
                               **kwargs)
         
-    def write_results(output_basename: str):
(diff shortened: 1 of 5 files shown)


thjsal:
Looks like you still import os here, even though you removed os from the UltraNest imports:

https://github.com/xpsi-group/xpsi/blob/b59412433b7c9f734d0f45e4256144e22e9ed6c8/xpsi/UltranestSampler.py#L54

This makes the CI tests to fail.

MHoogkamer:
Change saving as npy to hdf5, because npy files are dependent on the numpy version you have. This might cause issues in the future

PR #557, opened 2025-05-13
Added a model loader and updated the Module_generator tutorial to present it

### LINKED ISSUE #481: Add a model loader ?
If you load only a single main file in a notebook without a config file, then everything is fine. But as soon as there are config files and/or more than 1 model, it gets very messy easily. So far, a work around to make it not too hard has been to name every files differently (not only the main but also all the config and the custom classes it depends on) if we wanted to load them. This is bothering to make such changes after the run is finished.

I propose to make a module loader, which will load any main with an associated config file. It will then clear the cache so other models can be loaded safely.

lmauviard:
Closes #481


sguillot:
I have 2 suggestions:
- give the documentation for the model_loader elsewhere than in the module_generator, I'm not sure where. Maybe a page by itself ?
- With the documentation of the model_loader, I suggest adding info on how to use it for multiple models (which is its main purpose). Something like:
```
model_name = 'CST_PDT'
config_names = [ '4kLP_LR_seed0',  '4kLP_LR_seed1'  ]
runs = {'4kLP_LR_seed0':['run0'],   '4kLP_LR_seed1':['run1'] }
# Note that you cannot have two runs with the same name (key and value)
all_models = {}
for config in config_names:
    all_models[config] = load_model( model_name , config_name = config )
    ```

sguillot:
everything works.

But I also suggest to add documentation on the use of the `execution path` option

ckazantsev (commented):
if no config_path, `os.path.abspath(config_path)` is not happy about the NoneType and causes an error

ckazantsev (commented):
but if i put my model path as config path it works without (apparent) problems

ckazantsev (commented):
looks good to me now

lmauviard:
The requested fixes have been made ! Can anyone check the updated PostProcessing notebook ? @sguillot ?

sguillot (approved):

pstammler:
It worked for me too.

pstammler (approved):

PR #560, opened 2025-05-13
Update Data.py

sguillot:
Simply adding a check that the number of `phases` is compatible with the number of columns of `counts`)


thjsal (approved):
Looks good to me.

lmauviard (approved):
Looks good to me too

PR #561, opened 2025-05-13
551-choice-of-customization-inside-module-generator

pstammler:
This module-generator script includes potential structures for modules if the observational data are stored into fits files.


xpsi/module_generator.py
denis-gc, line 1: The changes looks good.

denis-gc (approved):
The changes looks good.

PR #562, opened 2025-05-13
Multiple imaging tutorial update

thjsal:
A shortened and cleaned Multiple Imaging tutorial brought back to the documentation pages. Could probably simplify it even more, but I think it is now more or less ok (unless people want changes).


drannawatts (approved):
Looks good, merge away!!

PR #567, opened 2025-05-13
rayXpanda removal

thjsal:
The linking to the rayXpanda package and the codes only used together with that package have been removed (e.g. the `compute_derivative() `function in `ray.pyx`). I tested that the X-PSI installation still works and test examples can be run.


sguillot:
There are still mentions of rayXpanda in 2 notebooks:
- global_surface_emission
- Multiple_imaging.ipynb

thjsal:
They are mentioned there in a "historical context" but not used. So I think it should not be an issue.

sguillot:
I see, I was looking at the old Multiple_Imaging notebook (which is still in this PR, and which add `xpsi.cellmesh.__deactivate_rayXpanda = True`. 
But I see that your other PR on Multiple Imaging does not have it anymore.

lmauviard (approved):
Looks good to me. I manage to install X-PSI and run the example_fast. There are still many occurrences of rayXpanda in the old Multiple Imaging Notebook but I guess this is updated in PR#562. Otherwise, all other occurrences are clearly removed.

PR #571, opened 2025-05-14
Update _corner.py

sguillot:


thjsal (approved):
Looks good and works for me too.

ckazantsev (approved):

PR #580, opened 2025-05-14
Checking that T and logg within atmosphere bounds if using nsx

thjsal:
Added a few checks before starting photosphere integration that the given parameters are within atmosphere table limits. Currently, this assumes the atmosphere table has the nsx format. But it is possible to customize this. I guess only issue is that this check needs to be done on every pulse profile computation. The reason is that we do not know in advance whether the surface gravity for the sampled mass and radius will be within the atmosphere bounds. Similar check is also done typically in the call-function of CustomPrior, but the check here can be an additional safety measure since it does not require user to manually set the surface gravity limits. Or if we modify the standard CustomPrior to also automatically obtain the surface gravity limits, then the checks here may become not necessary anymore.


sguillot (approved):
All good!

PR #584, opened 2025-05-15
Cython double==double comparisons made approximate

thjsal:
All the double == double comparisons that I found (at least in the Cython codes) were replaced with `are_equal()` function that I placed in `tools/core.pyx` and that allows 1e-12 tolerance (by default) around the exact value.

I tested a couple of the test example runs, that they give still the same results as before.


sguillot:
@thjsal New modifications are ready for you to test

sguillot (approved):

thjsal:
Thanks! These seem to work. I will merge it now.

PR #588, opened 2025-05-15
Correct fits reading and support trimming

### LINKED ISSUE #534: Support is not trimmed in the Signal
The new method to include min and max channel relies on giving it in input of Signal constructor. This trims both the instrument and the data. However does not trim the support accordingly. This should be fixed

### LINKED ISSUE #574: Error when loading fits Instrument with TLMIN4 > 0
When using `from_ogip_fits` to load an Instrument, there is a bug when TLMIN4 > 0. This IS NOT the case for NICER and XMM responses, which both have TLMIN=0, so no problems on this end. 

However if you did load other responses with TLMIN4 > 0 and specifying a `max_channel` value different than -1, you have had your channels shifted by TLMIN4 unit(s). If you let `max_channel=-1`, this would have always resulted in X-PSI raising an error.

lmauviard:
This implements a few changes : 
Closes #534
Closes #574
Changed the way to trim response : Now the response is trimming relies on bounds being passed to the instrument during model convolution.

Moreover, there are no longer min/max inputs in the response and the min/max channels can only be provided for the data or the Signal. This is in principle backward compatible if the user provides the right min/max channels to the Data/Signal, as X-PSI now throws a warning if the user tries to provides these deprecated values to Instrument.

Still in progress as there seems to be a very little  discrepancy with previous implementation (in my test case, Delta LogL ~ 10e-5, only one value in response changes, and ~0.01% change in spectrum). I suspect the change in spectrum comes from the long tail of the NICER response, which is now not trimmed and implies more sparse values of energies on which the pulse profile is computed for the same `num_energies`


lmauviard:
@thjsal @sguillot @yveskini I implemented a method to trim the matrix input energy automatically using the cumulative of the response. Min/max input are deprecated and min/max channel should be used only in the Signal (or in the Data but preferably in the Signal)
Can you test if this works for you/with your implementations ? Everything should be fixed now, I will add some more specification in the doc. We have to make sure the user does not overwrite the Instrument.__call__() method, otherwise this creates a bug and I do not see any way the user would need to modify it. Maybe with the pile-up @ckazantsev ?

ckazantsev:
Pileup does require a different Instrument.call() (see the InstrumentPileup function at the end of the Instrument.py)... as it needs to be implemented between ARF and RMF, `np.dot(matrix,signal)` is not relevant

lmauviard:
Since this PR has been hanging around for a while (since Hackweek !), I added a few very little tweaks to it no related to the fixed issue. Can someone have a look at it all and finally close it ?

thjsal:
Looks like this upgrade is not backward incompatible, because it breaks the current X-PSI 101, and probably also at least the Modeling tutorial. I get this error:
`TypeError: Instrument.trim_response() got an unexpected keyword argument 'threshold'`

I think you should update those tutorials and warn in the changelog the old "threshold" argument was removed (and is thus not working anymore).

xpsi/Instrument.py
```diff
@@ -353,7 +355,7 @@ def trim_response(self,
             cumsum[i] /= cumsum[i,-1]
 
         # Extract with the adapted percentile
-        under_tolerance = [ _np.all( cumsum[:,i] <= 1 - tolerance ) for i in range( cumsum.shape[1] ) ]
+        under_tolerance = [ _np.any( cumsum[:,i] <= 1 - tolerance ) for i in range( cumsum.shape[1] ) ]
```
thjsal: Are you sure this change (from your last commit) keeps the original logic (mentioned in the Changelog information) that the 
> cumulative of all the channels to reach the (1-tolerance) quantile
?

Isn't this code now requiring that at least one channel (but not necesserily more than that) is below the tolerance?

thjsal: If you decided to change the logic, I think the docstring and changelog information should be still updated.

PR #589, opened 2025-05-15
Renaming samplers classes and methods

sguillot:


sguillot:
Thank you @pstammler for the tests.
But I reverted your commits to the previous one, because the changes you made (Instrument TXT -> FITS) should really be done in another issue and pull request (to avoid mixing things up)

thjsal (approved):
Looks good and I tested that it works for me too.

pstammler (approved):

Issue #570, opened 2025-05-14
use_nestcheck=True  breaks the post-processing


sguillot:

sguillot:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
Cell In[25], line 2
      1 # Loading the run again :)
----> 2 ST.runs = xpsi.Runs.load_runs(ID='ST',
      3                                run_IDs=['run'],
      4                                roots=['ST_live_1000_eff_0.3_seed0'],
      5                                base_dirs=['../../examples/examples_fast/Outputs/'],
      6                                use_nestcheck=[True],
      7                                kde_settings=getdist_kde_settings,
      8                                likelihood=ST.likelihood,
      9                                names=ST.names,
     10                                bounds=ST.bounds,
     11                                labels=ST.labels,
     12                                truths=ST.truths, # Adding this line
     13                                implementation='multinest',
     14                                overwrite_transformed=True)

File ~/Softwares/anaconda3/envs/xpsi/lib/python3.11/site-packages/xpsi-3.0.6-py3.11-macosx-11.1-arm64.egg/xpsi/PostProcessing/_runs.py:191, in Runs.load_runs(cls, ID, run_IDs, roots, base_dirs, use_nestcheck, likelihood, multi_mode, mode_label, **kwargs)
    187 runs = []
    188 for root, run_ID, base_dir, check in zip(roots, run_IDs,
    189                                          base_dirs,
    190                                          use_nestcheck):
--> 191     runs.append(NestedBackend(root, base_dir,
    192                               ID=run_ID,
    193                               use_nestcheck=check,
    194                               transform=_transform,
    195                               overwrite_transformed=_overwrite,
    196                               **kwargs))
    197 Runs = cls(runs, likelihood, ID, **kwargs)
    198 Runs.multi_mode = multi_mode

File ~/Softwares/anaconda3/envs/xpsi/lib/python3.11/site-packages/xpsi-3.0.6-py3.11-macosx-11.1-arm64.egg/xpsi/PostProcessing/_backends.py:122, in NestedBackend.__init__(self, root, base_dir, use_nestcheck, transform, overwrite_transformed, **kwargs)
    120 if implementation == 'multinest':
    121     try:
--> 122         self._nc_bcknd = process_multinest_run(dead_points, live_points, root, base_dir=base_dir)
    123     except FileNotFoundError:
[...]

PR #590, opened 2025-05-15
Postprocessing tests

thjsal:
Unit and CI tests added for post-processing routines. Also fixed a couple of deprecation warnings from signalplotter and raised a better error message for background plotter if having only one sample when trying plot the uncertainty range.


thjsal:
I needed to change some syntax in Projection Tool so that Py3.9 test also passes (some code added there by @lmauviard ). 

Now the tests should not pass anymore, since I switched `use_nestcheck=True`, which leads to the bug that has not been fixed yet, except apparently in this pull request: #537 .

MHoogkamer:
Issue #570 is fixed in pull request #591

MHoogkamer:
Looks good to me!

MHoogkamer (approved):

PR #597, opened 2025-05-16
Postprocessing test updated with module loader

thjsal:
Updated the new post-processing GitHub tests to use the new module loader


sguillot (approved):
All good

PR #601, opened 2025-05-16
596 unit tests for interstellar.py

sguillot:
First attempt at unit tests for the interstellar module.
Do they make sense @dhuppenkothen ?

### DIFF
--- a/xpsi/tests/test_interstellar.py
+++ b/xpsi/tests/test_interstellar.py
@@ -1,0 +1,100 @@
+import unittest
+import numpy as np
+from xpsi.Interstellar import Interstellar
+
+class TestInterstellar(Interstellar):
+    def attenuation(self, energies):
+        # Mock implementation for testing
+        return np.ones_like(energies)
+
+class TestInterstellarMethods(unittest.TestCase):
+    def setUp(self):
+        self.model = TestInterstellar()
+
+    def test_attenuation(self):
+        energies = np.array([1, 2, 3])
+        expected = np.array([1, 1, 1])
+        result = self.model.attenuation(energies)
+        np.testing.assert_array_equal(result, expected)
+
+    def test_call_with_1d_signal(self):
+        energies = np.array([1, 2, 3])
+        signal = np.array([10, 20, 30])
+        self.model(energies, signal)
+        expected_signal = np.array([10, 20, 30])
+        np.testing.assert_array_equal(signal, expected_signal)
+
+    def test_call_with_2d_signal(self):
+        energies = np.array([1, 2, 3])
+        signal = np.array([[10, 20], [30, 40], [50, 60]])
+        self.model(energies, signal)
+        expected_signal = np.array([[10, 20], [30, 40], [50, 60]])
+        np.testing.assert_array_equal(signal, expected_signal)
+
+    def test_invalid_signal_type(self):
+        energies = np.array([1, 2, 3])
+        with self.assertRaises(TypeError):
+            self.model(energies, list(range(10)))  # Passing a list instead of ndarray
+
+    def test_invalid_signal_dimensions(self):
+        energies = np.array([1, 2, 3])
+        with self.assertRaises(ValueError):
+            self.model(energies, np.array([[1]]))  # Invalid dimensions
+
+if __name__ == '__main__':
+    unittest.main()
+
+
+#
+# # File: xpsi/tests/test_Interstellar.py
+# import numpy as np
+# import pytest
+# from xpsi.Interstellar import Interstellar
+#
+#
+# class MockInterstellar(Interstellar):
+#     """ Mock class implementing the abstract attenuation method. """
+#
+#     def attenuation(self, energies):
+#         return np.exp(-0.1 * energies)
+#
+#
+# class TestInterstellar:
+#     def test_signal_dimension_one(self):
+#         energies = np.array([0.1, 0.5, 1.0, 2.0])
+#         signal = np.array([10.0, 20.0, 30.0, 40.0])
+#         interstellar = MockInterstellar()
+#         interstellar(energies, signal)
+#         expected_signal = signal * np.exp(-0.1 * energies)
+#         print(signal)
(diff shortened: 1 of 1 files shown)


dhuppenkothen (commented):
Curious why these classes are Nested?

xpsi/tests/test_interstellar.py
```diff
@@ -0,0 +1,56 @@
+import numpy as np
+import pytest
+from xpsi.Interstellar import Interstellar
+
+class TestInterstellar:
+
+    class MockInterstellar(Interstellar):
+        def attenuation(self, energies):
+            return np.exp(-energies)  # Mock implementation for testing
+
```
dhuppenkothen, line 10: Curious why these are nested classes?

xpsi/tests/test_interstellar.py
```diff
@@ -0,0 +1,56 @@
...
+    def setup_method(self):
+        self.model = self.MockInterstellar()
```
dhuppenkothen, line 12: Unless something has changed in PyTest (and it might well have), I'd expect there to be a method called `setup_class` and anything you'll want to use later on to be stored in `cls.model`.

xpsi/tests/test_interstellar.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def test_attenuation(self):
+        energies = np.array([1., 2., 3.])
+        expected = np.exp(-energies)
+        result = self.model.attenuation(energies)
+        np.testing.assert_allclose(result, expected, rtol=1e-5)
```
dhuppenkothen, line 18: In principle this test makes sense, but the nitpick I'd have with it is that it tests your *mock class*, which you defined as part of the tests, so it doesn't really test the `Interstellar` class, does it? Or maybe there's stuff happening in the background I'm not aware of?

xpsi/tests/test_interstellar.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def test_call_single_dimension(self):
+        energies = np.array([1., 2., 3.])
+        signal = np.array([10., 20., 30.])
+        expected_signal = signal * np.exp(-energies)
+        self.model(energies, signal)
+        np.testing.assert_array_equal(signal, expected_signal)
```
dhuppenkothen, line 25: I'm not sure I understand what this test is doing: `signal` and `expected_signal` are both defined in the test, so what the assertion is testing is entirely code inside this test, and not related to X-PSI at all, as far as I can tell. 

Does `self.model()` return anything? I'm assuming that's what you actually want to compare to?

Also, this uses `np.testing`, and I'm not sure whether numpy's testing framework will clash with pytest. Could this just be an assertion, for example with `np.isclose()`?

xpsi/tests/test_interstellar.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def test_call_two_dimensions(self):
+        energies = np.array([1., 2., 3.])
+        signal = np.array([[10., 15.], [20., 25.], [30., 35.]])
+        expected_signal = signal * np.exp(-energies[:, np.newaxis])
+        self.model(energies, signal)
+        np.testing.assert_array_equal(signal, expected_signal)
```
dhuppenkothen, line 32: Same issue as test above?

xpsi/tests/test_interstellar.py
```diff
@@ -0,0 +1,56 @@
...
+        energies = np.array([1., 2., 3.])
+        signal = 'invalid'
+        with pytest.raises(TypeError, match='Signal must be a numpy.ndarray.'):
+            self.model(energies, signal)
+
+    def test_call_invalid_signal_type_list(self):
+        energies = np.array([1., 2., 3.])
+        signal = [10., 20., 30.]  # List instead of ndarray
+        with pytest.raises(TypeError, match='Signal must be a numpy.ndarray.'):
+            self.model(energies, signal)
```
dhuppenkothen, line 44: Do the inputs actually *have* to be `numpy.ndarray`s, or could they generically be iterables? Is there a good reason why `signal` *can't* be a list? Or a `torch.tensor`? Or something else that stores vectors?

xpsi/tests/test_interstellar.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def test_call_invalid_signal_dimensions(self):
+        energies = np.array([1., 2., 3.])
+        signal = np.array([[10., 15.]])  # Invalid shape
+        with pytest.raises(ValueError, match="non-broadcastable output operand with shape"): # with shape (1,) doesn't match the broadcast shape (3,)"):
+            self.model(energies, signal)
```
dhuppenkothen, line 50: This should fail in two separate ways, that probably should be tested in separate tests. First, here this test is written to fail because the dimensionality of `energies` doesn't match with the dimensionality of `signal` . However, I think this should probably also fail because `energies` is of length 3 and `signal` is of length 2. Or can you have energies and signals that are of unequal length? I think not, at least not in your mock class. Again, you might run into some trouble here because you're mainly testing the mock class, not actual functionality in the superclass, unless there's stuff I can't see?

xpsi/tests/test_interstellar.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def test_call_invalid_signal_dimensions3D(self):
+        energies = np.array([1., 2., 3.])
+        signal = np.array([[[10., 15.]]])  # Invalid shape
+        with pytest.raises(ValueError, match="Invalid number of signal array dimensions"):
+            self.model(energies, signal)
```
dhuppenkothen, line 56: This test is basically testing the same as the above, as far as I can tell?

dhuppenkothen (changes requested):
It's a great start, but I think your AI assistant might need a bit more practice 😉

xpsi/tests/test_interstellar.py
thjsal: I think `self.model()` modifies the "signal" internally as far as I remember. So the signal variable is different after you called that.

PR #614, opened 2025-08-17
Likelihood fixes

thjsal:
The Poisson likelihood with a fixed background was modified so that it can have the correct normalization if the user provides the precomputed ln-data-factorial term (as usually done for the default background marginalized likelihoods). So far this term has been ignored if fixing the background.

In addition, the normalization of the Gaussian likelihood functions (used for polarization analysis) was fixed to have the correct sign.


thjsal:
I sneaked in also a correction for the typo present in the comments of all integrators (that I mentioned in Slack to Bas and Yinghan).

 @Basdorsman Are you also ok if merge this pull request? You should still be able to obtain the old likelihood values using the Poisson extension if not giving the normalization factor as an input parameter. So it should not break your old scripts.

PR #615, opened 2025-08-20
Snellius instructions updated

thjsal:
I updated the suggested partition name for the Snellius example script, and added a couple of python packages (usually/often needed) to the Snellius installation instructions.


yveskini:
@thjsal , I think the astropy should  version should be between 5.2 and 7.0.0,
So it should be: `pip install "astropy>=5.2,<7.0.0"`

thjsal:
But sure, we can add the restriction there if we just want to be consistent with the "normal instructions".

PR #621, opened 2025-11-04
Fixed nestcheck behaviour for multimode postprocessing

### LINKED ISSUE #617: use_nestcheck=True for multimodal sampling
When you apply nestcheck in the PostProcessing on runs from multimodal nested sampling analyses, you can have live points coming from one converged mode that present loglikelihoods that are lower than those from dead points of another mode, which can cause an error that breaks the loading.

In _nestcheck_modifications.py : 
(Line 73-77)
dead = dead[:, :-2]
live = live[:, :-1]
assert dead[:, -2].max() < live[:, -2].min(), (
    'final live points should have greater logls than any dead point!',
    dead, live)

lmauviard:
When loading runs for post-processing, ``use_nestcheck=[True]`` did to not work when ``multi_mode=True`` was enabled. This has been fixed.

Closes #617

### DIFF
--- a/changelog.d/621.attribution.rst
+++ b/changelog.d/621.attribution.rst
@@ -1,0 +1,1 @@
+Lucien Mauviard


DevarshiChoudhury (approved):
LGTM! 👍🏽

PR #622, opened 2025-11-05
620 custom interstellar extrapolate energy range

### LINKED ISSUE #620: CustomInterstellar extrapolate energy range
The Custom Interstellar uses interpolator to get the base attenuation values at the energies of the ``input`` of the response. These base attenuation values are then put to the power of NH to get the actual attenuation. The base attenuation table that has been used until now covers 0.1-12 keV range, so any response with energies outside this range will need extrapolated values from the table.

In particular, the attenuation values around 12 keV are close to 1 and they get extrapolated to values that are larger than 1 above 12 keV. As the attenuation varies exponentially with NH, when NH is high ($\sim 5 \times 10^{20} \text{cm}^{-2}$ in my case) it explodes to very high values. If the instrument has some, even tiny, redistribution at these energies, it will become a significant contribution to the flux. This is the case with NICER if you load the whole response without trimming it.

This does not affect any prior NICER analysis as the maximum input has been set for these, but an unaware user may make this mistake.

To make sure this does not happen, a possible fix would be to use a 0.01-20 keV table (works for responses of NICER, XMM and Chandra). We could also add a check that the base attenuation values are not extrapolated, or that the attenuation values are all smaller than 1.

lmauviard:
2 fixes for issue #620 are provided. 

The first is in the general `Signal.py` class, where it is checked whether the Custom interstellar has attenuation greater than one. This works for all Custom Interstellar implementations, although it does not test the whole range of possible parameters, this is only a basic check a the beginning of the run. This should work for all currently existing CustomInterstellar implementations.

The second is in `module_generator.py`, where a condition on the maximum limit value of the attenuation is provided for the `CustomInterstellar.py` that is generated. I think this will make the code a bit slower during inference, but any user can remove it if confident that this will not happen in his case. 

@thjsal do you think we need to set `self._interpolator.extrapolate` to `False` just to be sure or not ?

Closes #620


thjsal:
> @thjsal do you think we need to set self._interpolator.extrapolate to False just to be sure or not ?_

I am not sure if it is needed. Either way is fine for me.

PR #628, opened 2025-12-10
Correcting mistakes in new `InstrumentPileup`

ckazantsev:
the call for super() was missing parameters and leading to an issue


thjsal:
In my opinion having 3.2.1 would make it clear that from now on we are using X-PSI version where this bug is fixed.

PR #631, opened 2025-12-23
Adding different emission models

lmauviard:
I implemented a framework to add any model to X-PSI, even ones that would need to be embedded in the metric. Such models can have parameters that are fitted alongside the usual models. 

Alongside this, I provide a Powerlaw model, that can be pulsed or not. I managed to fit some data with this model. I can maybe add some other basics models if needed, such as a Gaussian or different pulsation definitions ?

Should I make a new notebook to explain it, or simply add a section to the Modeling/XPSI101 notebooks ?

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -39,6 +39,7 @@
    Post-processing
    Emitting_patterns_2Dprojection
    Accretion_disk
+   Other_Emission_Models
    Multiple_imaging
    Importance_sampling
    Module_generator_tutorial


docs/source/Other_Emission_Models.ipynb
thjsal: Comments regarding the new Other Emission Models -notebook:

1. I suggest to re-run this notebook so that it will print the new version number "3.3.0" in the beginning.

2. In the sentence " In this case it is important to use this likelihood as the likelihood with background maginalization with result in wrong parameter inference.", the latter "with" should perhaps be "would"?

3. "To use the likelihood with background maginalization you need to have some background constraints." Wasn't this what was done e.g. in Riley+2019? Or do you mean that having background constraints is extra important when having a power-law model?

TBD: Mode comments may still come ....

xpsi/EmissionModels.py
```diff
@@ -0,0 +1,249 @@
...
+                symbol = r'$\gamma$',
+                value = values.get('gamma', None))
+        
+        # Add more if pulsed
+        self.is_pulsed = pulsed
+        if self.is_pulsed:
+
+            # Define oscillation amplitude parameter
+            oscillation_amplitude = Parameter('oscillation_amplitude',
+                    strict_bounds = (0. , _np.infty),
```
thjsal: An oscillation amplitude larger than 1 means that the flux can become negative. Is this something we want to support?

thjsal:
I just re-ran the notebook to fix the oscillation amplitude strict upper bound from infinity to 1 also in the Gaussian example. Otherwise it looks all good to me. I just remind that the updated module generator provided by @pstammler has not been implemented yet (but for me it is ok to add later if wanted). 

> Currently, a phase independent signal need to be computed for >=5 phases so the energy_integrator in the signal register does not bug. There are two possible things to do to handle this:
> 
> Add an if statement in the signal register, but this will affect the computation time of all likelihoods
> Compute 5 times the same thing to be able to apply the energy_integrator, which make using the emission model slower.
> Does anyone have an opinion on this ? I think I prefer option 2


I think the option 2 is fine. It does not at least make the whole phase-independent signal computation 5 times slower, right?

PR #684, opened 2026-03-06
Fixing notebook layout

sguillot:


sguillot:
This is all fixed now. I compiled locally to double-check.

Screenshot 2026-03-06 at 14 23 01

Issue #641, opened 2026-03-02
Energy integration to review?


pstammler:
In the pull request draft 624, there is another approach proposed for energy integration in case of discontinuous instrumental responses. That changes the way to define the energy edges. Perhaps it would be interesting to improve and adapt it for all instruments...

PR #691, opened 2026-03-11
#641 2D channel and energy edges, with filtering options

pstammler:
In this draft, a new structure is proposed for the extracted energies and channels. New features include the use of some datasets statistical errors and optional filtering regarding the quality of the data or the grouping of channels.

The approach for the grouping, however, is not complete yet. For now, it is working with channels grouped to get a minimum number of counts. Grouping with binning factors shall be added in future releases...


thjsal (changes requested):
I did not look yet into all the details, but it seems clear that this a currently backward-incompatible update. If keeping it like this, we should do a major version bump to X-PSI v4. Or modify all the functions so that they are backward-compatible.

sguillot:
To be tested:

- [ ]  main.py
- [ ]  TestRun_BB.py
- [ ]  TestRun_Num.py
- [ ]  TestRun_NumBeam.py
- [ ]  TestRun_Pol.py
- [ ]  TestRun_PolNum_split_1spot.py
- [ ]  TestRun_PolNum_split_inference.py
- [ ]  TestRun_AMXP.py
- [ ]  Modelling Notebook
- [ ]  Synthesis Notebook
- [ ]  Instrument Notebook

PR #695, opened 2026-03-15
Fixed corrupted Instrument Synergy

sguillot:
reverting to older version, removing do_fast option, and re-running all


thjsal:
Just to mention: When opening the notebook I get also this warning:
`Notebook validation failed: Additional properties are not allowed ('jetTransient' was unexpected)`.

However, I get this same warning for some other notebooks as well. I guess someone created these with newer packages than what I am using, and thus my Jupyter cannot understand the syntax.

thjsal (approved):
I will approve it now, but as I commented above, I am not sure if the figures will be displayed correctly in the docs. But let's just test it.

PR #703, opened 2026-03-20
Adding .zenodo.json file for Zenodo integration

### LINKED ISSUE #702: Add .zenodo.json file to fix Zenodo Integration
New releases have stopped appearing on Zenodo [https://doi.org/10.5281/zenodo.7632628](https://doi.org/10.5281/zenodo.7632628) - for the last two releases (v3.2.0 and v3.3.0) we have had an empty and hence uninformative error message. 

We previously thought this was due to the issues with the Github actions which were fixed in the v3.3 release, but we now suspect it may be because we do not have a .zenodo.json file (which we need since the project is listed there as part of the EU Open Research Repository). See discussion here:  [[https://support.zenodo.org/help/en-gb/24-github-integration/204-why-does-my-github-release-fail-with-an-error](https://support.zenodo.org/help/en-gb/24-github-integration/204-why-does-my-github-release-fail-with-an-error)]

drannawatts:
Would close #702 
Hopefully contains sufficient information (in the right format) to permit Zenodo to update automatically on releases again.

### DIFF
--- a/.zenodo.json
+++ b/.zenodo.json
@@ -1,0 +1,155 @@
+{
+    "creators": [
+        {
+            "name": “Riley, Thomas E.”,
+            "orcid": "0000-0002-1825-0097"
+        },
+        {
+            "name": “Salmi, Tuomo”,
+            "affiliation": "University of Helsinki",
+            "orcid": "0000-0001-6356-125X"
+        },
+        {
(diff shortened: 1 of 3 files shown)


thjsal (approved):
Let's see if it works!

PR #708, opened 2026-04-21
Update HPC_systems.rst

sguillot:


sguillot:
This PR will be deprecated soon, since CALMIP has a new cluster, which will require a different setup!

sguillot:
@pstammler @lmauviard @ckazantsev 
The installation for Kairos now works (I think).  I've ran an example script which ran multinest correctly.
My first job submission is still pending, so I don't have a test of a multi-node run.

Please test the installation, and if everything works, you can approve the PR.

ckazantsev:
The installation worked perfectly for me. Will test launching a job later today just to confirm, but it looks good to me

ckazantsev (approved):

PR #709, opened 2026-05-26
Fixed the backscale for signal synthesization

lmauviard:
Previously misused the input `data_BKG` in the `Signal.synthesize` method, which resulted in overestimation of the background in the generated expected counts. 

This is now fixed and `data_BKG` is actually the data counts spectrum, that will be rescaled by the backscal ratio.

### DIFF
--- a/changelog.d/709.fixed.rst
+++ b/changelog.d/709.fixed.rst
@@ -1,0 +1,1 @@
+Fixed the Signal.synthesize method when using data_BKG, which previously resulted in overestimated background contribution.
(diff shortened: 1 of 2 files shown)


sguillot:
Do we need to also update the synthetic data notebook ?

sguillot (approved):

lmauviard:
Not needed since we used `backscal_ratio=1.0` in the notebook, so it worked ! But maybe I can change so it uses `backscal_ratio != 1.0` instead ?

sguillot:
I guess it's fine since the notebook should work universally whether the back_scale is 1 or something else

PR #712, opened 2026-06-22
XPSI numpy 1.x to 2.x conversion

dhuppenkothen:
Okay, so I audited the code with the help of Claude, and it doesn't look like there's a lot that (should?) break between numpy 1.x and 2.x in the X-PSI code base. 

All I did was update some references from numpy.infty to numpy.inf, and the tests all run fine, except for anything that relies on MultiNest, which I didn't install.

I'll make a PR here and see what the continuous integration has to say. 

It'd be good if someone else (who uses X-PSI more than I do) could try out this branch on a couple of test cases and report back.


thjsal:
I tested this branch and found that the **TestRun_BB.py** (in _examples/examples_modeling_tutorial/_) crashes with the following error message:

```
... 
Starting MultiNest
 generating live points
ERROR in loglikelihood: only 0-dimensional arrays can be converted to Python scalars
Exception ignored while calling ctypes callback function <function run.<locals>.loglike at 0x752e1c5ff5e0>:
  File "/home/saltuomo/.conda/envs/xpsi/lib/python3.14/site-packages/pymultinest/run.py", line 223, in loglike
    return LogLikelihood(cube, ndim, nparams, lnew)
  File "/home/saltuomo/.conda/envs/xpsi/lib/python3.14/site-packages/pymultinest/solve.py", line 67, in SafeLoglikelihood
    sys.exit(1)
SystemExit: 1
```
Perhaps it indicates that updates are needed on pymultinest too due to NumPy upgrade (github actions are not testing for it yet, as we did not yet manage to make GitHub to install MultiNest)? Interestingly though, in the simplest x-psi fast example (_examples/examples_fast/Modules/main.py_) pymultinest works still fine.

thjsal:
Okay, I found the mistake was actually coming from X-PSI itself (Likelihood.py) and fixed it. This part was not used in the simplest example and therefore it caused no errors there. Now all test runs I have tried seem to work.

sguillot:
I think there is another conflict with numpy>2.0 and matplotlib.
```
matplotlib 3.7.3 requires numpy<2,>=1.20, but you have numpy 2.5.1 which is incompatible.
```

And higher version of matplotlib give the following error in the post-processing:
```
AttributeError: module 'matplotlib.cm' has no attribute 'get_cmap'
```

sguillot:
In addition to the tests of the `example runs` done by @thjsal, I did additional tests:
- generate synthetic data with both versions of X-PSI (the one with numpy 1.26 and the one with numpy 2.5) to confirm that they are identical (for the same seed)
- run inference with with both versions of X-PSI to confirm that the posteriors are similar (small differences come from the sampling)
- run full a Post-Processing notebook to confirm that everything works.

This confirms that the writing/reading from FITS files (including ARF/RMF) was not altered, and that the sampling hasn't changed as well.

So, I am validating this PR which could be merged.
And I would argue that there should be a release of X-PSI 3.4.0 with this update to numpy 2.0

thjsal:
Before merging we should still wait for fgivenx to accept Daniela's pull request or start using our own branch of it (and add the corresponding postprocessing unit test back).

sguillot:
What about temporarily changing to `use_fgivenx=False` in the `test_pulse_plotting_works` ?
If this is just for a unit test :-)

thjsal:
This is also used in the pulse and spectral plots in the Postprocessing tutorial (and this also the reason why this unit test exists). So that notebook should then also be fixed.

thjsal:
I wonder though how did you manage to run the Postprocessing tutorial (I see you said so above). Were you using the non-standard fgivenx version there?

sguillot:
Ah, that's a very good point.  I'll check what I did to made it work :-)

sguillot:
@thjsal  Indeed, my PostProcessing `PulsePlot` were using `use_fgivenx=False` (it does crash if I set it to True)
Honestly, the simplest approach might be to do this on the PostProcessing notebook as well.
I don't know what difference this makes to the plot.

thjsal:
Here is shown the difference between `use_fgivenx=True` (above) and `use_fgivenx=False` (below)
Screenshot from 2026-08-02 11-09-36
Screenshot from 2026-08-02 11-09-17

So in the latter no conditional-posterior contours are calculated and shown (so no different contour levels), but many individual posterior-sample curves are plotted on top of one another (as far as I understand).

Visually these two figures look similar (at least in this case), so I am fine with stopping the support for these contours and fgivenx if that is the easiest solution.

thjsal:
Or if we want to keep the conditional posterior bands, we should use modified fgivenx or wait them to approve the pull request.

sguillot:
Ok, I suggest to wait for the end of August, so that we can decide altogether what to do. That will also give more time to the FGIVENX people to approche Daniela's PR.

PR #717, opened 2026-07-16
Added correlation matrix plot in the post-processing

### LINKED ISSUE #716: Correlation matrix in post-processing

sguillot:
To run in a notebook, use:

```
matrix = xpsi.PostProcessing.CorrelationMatrix([ST.runs])
fig = matrix.plot(
    params=ST.names,
    IDs=OrderedDict([('ST', ['run',]),]),
    combine=False, combine_all=True, only_combined=False,
    method='pearson',        # or 'spearman' (rank; robust to non-linear degeneracies)
    triangle='lower',        # or 'full'
    cluster=False,           # True: reorder params so correlated blocks are adjacent
    threshold=None,          # e.g. 0.2: blank out |r| < 0.2
    annotate=True,
    write=False)
```


sguillot:
The output looks like this (for the Post-processing notebook example.
output

